PR #12, opened 2019-03-29
feat: add description to override supported fields

aaliakseyenka:
Description field should be in the list since ticket passes should contain dynamic texts

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -108,6 +108,7 @@ declare namespace Schema {
 
 	interface SupportedOptions {
 		serialNumber?: string,
+		description?: string,
 		userInfo?: Object | any[],
 		webServiceURL?: string,
 		authenticationToken?: string,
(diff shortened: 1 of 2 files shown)


aaliakseyenka:
@alexandercerutti please take a look on this. Would be great if you could include it in the nearest version

alexandercerutti:
Mhh, why should you be able to edit the description? Basically, if you have one type of pass, the description will be one. Moreover, if you have a localized pass, `description` is a Localized String. That mean that if you use `.localize` method, you can insert in description a placeholder that will be replaced in runtime by passkit with the description provided.

aaliakseyenka:
One type - event ticket for concerts. But we may have several tickets for different concerts. User may buy two tickets for two different concerts and based on Apple guidelines we need to show different description like one with date of first event and another with date of another event and also adding respective name of events

alexandercerutti:
Never heard this! Can you please link me that guideline?
Anyway, you are convincing me. 😁

alexandercerutti:
Published in v1.5.6! Thank you @aaliakseyenka

PR #13, opened 2019-04-01
fix: check both static list and remote list before raising error

aaliakseyenka:
@alexandercerutti please take a look on this MR. My case: I don't have anything in pass folder beside my `pass.json` file. All required icons I do using `load` API since icons are different for different events. In current master version I am getting error, although I specify all required data

### DIFF
--- a/src/pass.js
+++ b/src/pass.js
@@ -104,8 +104,9 @@ class Pass {
 			.then(([modelFileList, remoteFilesList, remoteBuffers]) => {
 				// list without dynamic components like manifest, signature or pass files (will be added later in the flow) and hidden files.
 				let noDynList = removeHidden(modelFileList).filter(f => !/(manifest|signature|pass)/i.test(f));
+				const hasAssets = noDynList.length || remoteFilesList.length;
 
-				if (!noDynList.length || ![...noDynList, ...remoteFilesList].some(f => f.toLowerCase().includes("icon"))) {
+				if (!hasAssets || ![...noDynList, ...remoteFilesList].some(f => f.toLowerCase().includes("icon"))) {
 					let eMessage = formatMessage("MODEL_UNINITIALIZED", path.parse(this.model).name);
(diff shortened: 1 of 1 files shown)


alexandercerutti:
Let me understand better. So you are downloading the icon every time you are creating a pass, based on the type of the pass. So you are saying that a check on remote files that were loaded is missing, aren't you? So if we don't check loaded files, if you don't have any files in your model except `pass.json` (which get excluded), you are getting `MODEL_UNINITIALIZED` error. Did I get the point?

aaliakseyenka:
you are right

alexandercerutti:
Perfect, I'll merge it asap, and release it in 1.5.8 (which is already ready to be released 👍). Then be sure to update to 1.6.0 as I'll release it: I'm going to add a planty of new fields introduced in iOS 12. 😉

aaliakseyenka:
so technically, I would like to have better control on what can be loaded dynamically and what should be static.
Also, question: do you handle the case when we have icon.png in static and also trying to load it dynamically, but call fails. Do you have this fallback to use static instead of dynamic?

aaliakseyenka:
>I'm going to add a planty of new fields introduced in iOS 12. 
do you have a link, describing new field? :)

alexandercerutti:
About the fields, here you are:
https://developer.apple.com/documentation/passkit/wallet/supporting_semantic_tags_in_wallet_passes

Also there is one more called "row" valid only for auxiliaryFields, but it is undocumented yet (found in wwdc video).

About the fallback, if I remember correctly, the downloaded medias have priority respect static medias. But if the download fails, it won't be inserted inside remote files/remote buffers.

alexandercerutti:
Published in [1.5.8](/alexandercerutti/passkit-generator/commit/1c8cf666fb340d66e4f53794a94f52f08f32a748) 🎉🎉🎉

Issue #21, opened 2019-05-04
logoText as override Option


gweit:
Hi Alex,

Is there any way to override the logoText?

As far as i can see it might work through the override-options, but is not declared in schema.js. Am i on the right path?

Best, Gernot

gweit:
i tried to add the following line in **schema.js** under **supportedOptions** and it works as expected:

**logoText: Joi.string()**

Could you please add it in one of your future updates?

alexandercerutti:
Hi @gweit,
yes, you are right, that's the correct way. But let me tell you first the reason I did not insert it in `supportedOptions`.
When I created `passkit-generator`, the idea behind was to have different models for different scopes. Therefore, when I was in the phase of attributes choosing, I filtered the ones that can, should or must be different for each pass from the ones that should be static for each model. This was a specific design, and it shouldn't be a problem to add it.
Tell me, do you really need it? Can you tell me your use case? Thank you.

gweit:
Hi @alexandercerutti,

Thanks for the fast reply.

You are right, and it won´t happen often that logoText will be changed - but as far as i am giving this option to the enduser it would be very cool to be able to change this one to ;)

alexandercerutti:
@gweit I'm definitely not sure about this thing. I mean... why should you be able to change the logoText? Usually, the logo text is the text of the... logo, or the company. For this reason, I asked you to show me your use case (or what the final pass should look like in different cases).

ianbale:
I have a use case where this is needed. UK Rail tickets have used the logo text field to display the travel date (no idea why ATOC make the decisions they do, but there is no possibility of getting them to change this). Therefore, it has to be different for every ticket. 

There is a work around, of course, simply create the JOSN file on the fly for every ticket, but since that has to be on DISK rather than in memory at the moment, that is not a great solution performance wise.

I've already got a patch to allow this to be set, along with backgroundColour (which I also need to override for some tickets).

Will get a branch setup with this shortly. Hopefully you'll accept the PR for this...

alexandercerutti:
What's ATOC? And which are the decisions you are talking about?
About the pass in memory, it will be available in the v2 (as per #14).
About the logoText, I left this issue open on purpose, to talk about the introduction of this edit. 👍

ianbale:
ATOC = Association of Train Operating Companies.

You really don't want to know! Money grabbing protectionists. Same everywhere...

alexandercerutti:
[...]

PR #28, opened 2019-05-31
Issue 21 set logo text

ianbale:
Re : #21 

Adds methods to permit setting of logoText and backgroundColour

### DIFF
--- a/src/fieldsArray.js
+++ b/src/fieldsArray.js
@@ -2,102 +2,76 @@ const schema = require("./schema");
 const debug = require("debug")("passkit:fields");
 
 /**
+ * Pass fields must be unique (for key) in its scope.
+ * Therefore we use a Set to keep them tracked.
+ */
+
+const fieldsKeys = new Set();
+
+/**
  * Class to represent lower-level keys pass fields
  * @see https://apple.co/2wkUBdh
  */
 
-class FieldsArray extends Set {
-  constructor(fields) {
-    super();
+class FieldsArray extends Array {
+	constructor(...items) {
+		super(...items);
+	}
 
-    /**
-     * Pass fields must be unique (for key) in its scope.
-     * Therefore we use a Set to keep them tracked.
-     */
+	/**
+	 * Like `Array.prototype.push` but will alter
+	 * also uniqueKeys set.
+	 */
 
-    this.fieldsKeys = new Set();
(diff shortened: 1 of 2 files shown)


src/fieldsArray.js
```diff
@@ -2,102 +2,76 @@ const schema = require("./schema");
 const debug = require("debug")("passkit:fields");
 
 /**
- * Class to represent lower-level keys pass fields
- * @see https://apple.co/2wkUBdh
+ * Pass fields must be unique (for key) in its scope.
+ * Therefore we use a Set to keep them tracked.
```
alexandercerutti: Please, keep the shortened address in the comment here. @ianbale

src/pass.js
```diff
@@ -284,6 +284,16 @@ class Pass {
...
+  logoText(text) {
+    this._props.logoText = text;
+    return this;
+  }
+
+  backgroundColor(colour) {
+    this._props.backgroundColor = colour;
+    return this;
+  }
+
```
alexandercerutti: There is no need to add methods for this. Also, by doing this, you should get anyway the errors, because the changes must be done in schema.js.
Also, `backgroundColor` is already settable through overrides.
Also, again, you cannot add only these two props. If you add methods for overrides, you should add them for ALL. For this reason, I decided to not create single methods, due to increased complexity in manteinance.

ianbale: I was unable to set backgroundColor through overrides. I get an error about it not being of type string... Hence created a method... Maybe a small change to overrides would fix that though?

alexandercerutti: That's weird. You should open a new issue for this so we can investigate.
Anyway, about logoText its the same: it should be enough to add it to schema. Do you mind if I do that? I have for sure another edit to apply (a different this). Moreover in this pull request you are bringing you edits to `fieldsArray.js`, which I'm not sure yet of.

ianbale: Yes. Please do. Just reject this PR and implement it the way you feel is best. You know the architecture far better than I have been able to work out for the sort time I've spent looking at it.

alexandercerutti: We have a deal! Will do ASAP. 👍

ianbale: Whatever my issues was with the backgroundColor the other day, I cannot replicate it now. I've also modified schema.js locally as you pointed out was the best option. Works perfect. So I'll await your update for that.

PR #31, opened 2019-06-06
Fixing passes parallel generation

### LINKED ISSUE #26: Cannot generate passes in parallel
Reference #25 

`fieldsArray.js` defined `fieldsKeys` outside of the `class`. This means they are shared across all class instances. 

Whilst the data is populated for a single instance, the static function `emptyUnique` clears this data for all instances.

The symptoms of this are that it is impossible to generate multiple passes in parallel (using promises).

ianbale:
fieldsArray.js
-----------------
I've added a wrapper class around the existing one. This will now hold the filedKeys set so that is it shared across all the array instances, but within an instance of the current pass. Other passes will have their own instance, and hence their own set of keys.

pass.js
----------
A small change required to the instantiation of fieldsArray and the way the data is copied into the pass data.

Closes #26

### DIFF
--- a/src/fieldsArray.js
+++ b/src/fieldsArray.js
@@ -6,16 +6,44 @@ const debug = require("debug")("passkit:fields");
  * Therefore we use a Set to keep them tracked.
  */
 
-const fieldsKeys = new Set();
-
 /**
  * Class to represent lower-level keys pass fields
  * @see https://apple.co/2wkUBdh
  */
 
-class FieldsArray extends Array {
-	constructor(...items) {
-		super(...items);
+class FieldsArray extends Object {
+	constructor(fields) {
+		super();
+
+		/**
+		 * Pass fields must be unique (for key) in its scope.
+		 * Therefore we use a Set to keep them tracked.
+		*/
+
+		this.fieldsKeys = new Set();
+		fields.forEach(a => this[a] = new ItemsArray(this));
+	}
+
+	addFieldKey(key) {
+		if (this.fieldsKeys.has(key)) {
+			debug(`Field with key "${key}" discarded: fields must be unique in pass scope.`);
+		} else {
+			this.fieldsKeys.add(key);
+		}
+	}
+
+	deleteFieldKey(key) {
+		this.fieldsKeys.delete(key);
+	}
+
+	emptyUnique() {
+		this.fieldsKeys.clear();
+	}
+}
+
+class ItemsArray extends Array {
+	constructor(owner) {
+		super();
 	}
 
 	/**
@@ -67,10 +95,6 @@ class FieldsArray extends Array {
 
 	get length() {
 		return this.length;
-	}
-
-	static emptyUnique() {
-		fieldsKeys.clear();
 	}
 }
 
--- a/src/fieldsArray.js
+++ b/src/fieldsArray.js
@@ -2,100 +2,102 @@ const schema = require("./schema");
 const debug = require("debug")("passkit:fields");
 
 /**
- * Pass fields must be unique (for key) in its scope.
- * Therefore we use a Set to keep them tracked.
- */
-
-/**
  * Class to represent lower-level keys pass fields
  * @see https://apple.co/2wkUBdh
  */
 
-class FieldsArray extends Object {
-	constructor(fields) {
-		super();
+class FieldsArray extends Set {
+  constructor(fields) {
+    super();
 
-		/**
-		 * Pass fields must be unique (for key) in its scope.
-		 * Therefore we use a Set to keep them tracked.
-		*/
+    /**
+     * Pass fields must be unique (for key) in its scope.
+     * Therefore we use a Set to keep them tracked.
+     */
 
-		this.fieldsKeys = new Set();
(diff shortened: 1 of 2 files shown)


ianbale:
This is ready for you to review. Sorry for the mix up earlier...

alexandercerutti:
Dude, you reintroduced spaces back in your code in the last commit 😂.
Anyway, I found the code flow really difficult to be read, basically to check if everything was okay.

So basically you are creating an "abstract class" to keep the changes.

I was thinking: what about instead saving the keys in the main class Pass?
I mean, the keys must be unique within the pass scope. So we might just remove the Set to be global in `fieldsArray.js`.

But this would open up to another problem: how do we link the field to the uniqueKeys?

I'd probably do something like this:

```javascript
    // In constructor
    this.fieldsKeys = new Set();
    fieldsName.forEach(name => {
        this[name] = new FieldArray(this.fieldKeys, ...arrayOfPossibleInitialValues);
    });

    // Inside fieldsArray.js
    class FieldArray extends Array {
        constructor(pool, ...args) {
            this.pool = pool;
            super(args);
        }

        ...
    }
``` 

So by doing this, we'd link the set into the Field and would be able to check at any operation on the Array.

Obviously, I thought this with v2 in mind. Let me know what do you think about 😉.
Anyway, we won't close this PR and the related issue before have found a solution to this problem.

src/fieldsArray.js
```diff
@@ -6,98 +6,65 @@ const debug = require("debug")("passkit:fields");
...
-
-  emptyUnique() {
-    this.fieldsKeys.clear();
-  }
+const poolSymbol = Symbol("pool");
+
+class FieldsArray extends Array {
+	
+	constructor(pool,...args) {
+    	super(...args);
```
alexandercerutti: @ianbale please look here, at the rows with wrong indentation: they are still using spaces. Just to let you know 😉

ianbale:
I like that! Simple and elegant...
I've tested it against some parallel pass generation and it works as expected.
I think I've eradicated the last few spaces...?

src/fieldsArray.js
```diff
@@ -6,98 +6,65 @@ const debug = require("debug")("passkit:fields");
...
-  }
-
-  get length() {
-    return this.length;
-  }
-}
-
-
-module.exports = FieldsArray;
+module.exports = FieldsArray;
```
alexandercerutti: @ianbale An empty end-line is always suggested to be inserted. Can you please insert it? Because if you won't do that, at my next edit or file opening, my editor will do that and I'll have to create a commit only for this or leave this as is (and causing a loop of auto-inserts, for me). At the end of the PR creation, I'll squash all your commits, so there won't be problems 😉

alexandercerutti:
> I like that! Simple and elegant...
> I've tested it against some parallel pass generation and it works as expected.
> I think I've eradicated the last few spaces...?

I'm glad you like it and it works 😄.
Yes, I made some comments on the last commit, so you can see what is left to be fixed. 👍

ianbale:
Just found out how to display the whitespace chars in Sublime text. Found two more lines with spaces.

alexandercerutti:
Great 😉 One more thing left and in the next few hours I'll push them (in the next release) 👍

ianbale:
one more thing?

alexandercerutti:
Yes, look at the [previous commit](https://github.com/alexandercerutti/passkit-generator/pull/31/commits/1204f19184eab55a63ed5a0bc13ea9cbfcd74c4f#r291498797). I've asked you to add the final line, because:

`if you won't do that, at my next edit or file opening, my editor will do that and I'll have to create a commit only for this or leave this as is (and causing a loop of auto-inserts, for me).`

Thank you 😊

alexandercerutti:
My bad!

ianbale:
Shall we call that quits after my screw up with the spaces? ;-)

alexandercerutti:
If I understood correctly what you mean, I suppose! 😂

Anyway, I think this is good enough to be merged 😄
I'll do it as soon as possible!

Thank you very much for your contribution. Would you mind to leave a star to the project? It is very important for me. 😊

ianbale:
Thank you for making the library available!

I've done loads of things that would be useful as libraries, but I just do not have the time available to wrap them up properly so that I can share them. I am well aware of just how much work is needed, so I really appreciate all those that do.

This has saved me a lot of time, so more than happy to give some of that saved time back to improve it for others.

src/fieldsArray.js
```diff
@@ -6,98 +6,65 @@ const debug = require("debug")("passkit:fields");
...
+	 * also uniqueKeys set.
+	 */
+
+	push(...fieldsData) {
+		const validFields = fieldsData.reduce((acc, current) => {
+			if (!(typeof current === "object") || !schema.isValid(current, "field")) {
+				return acc;
+			}
+
+			if (acc.some(e => e.key === current.key) || this[poolSymbol].has(current.key)) {
```
alexandercerutti: @ianbale I found right now that this logic is completely broken: first, the name of debug is `fieldsDebug`. Second, the field must be pushed into the accumulator only it the debug condition is false... 😢

PR #38, opened 2019-07-28
small fix of example in API

neogucky:
Noticed "pass.relevance("locations"," ...." was not working after copying it from the API example


alexandercerutti:
Oh, thank you very much! I'm merging this right now!

Just a note: please note that method `.relevance` is going to disappear in v2.0 😊

PR #46, opened 2019-12-04
Feature: support for overriding 'organizationName' field

WaveMeUp:
Hello!

First of all thanks for great library.
In our project we need to use different organization names so we have to override default pass.json property. Looks like your project doesn't have this ability.
We fixed it pretty easily with 'organizationName' property in 'OverridesSupportedOptions' interface and 'supportedOptions' object. 

Here is simple pull request, looking forward to your response!

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -66,6 +66,7 @@ const instance = Joi.object().keys({
 export interface OverridesSupportedOptions {
 	serialNumber?: string;
 	description?: string;
+	organizationName?: string;
 	userInfo?: Object | Array<any>;
 	webServiceURL?: string;
 	authenticationToken?: string;
@@ -82,6 +83,7 @@ export interface OverridesSupportedOptions {
 const supportedOptions = Joi.object().keys({
 	serialNumber: Joi.string(),
 	description: Joi.string(),
+	organizationName: Joi.string(),
 	userInfo: Joi.alternatives(Joi.object().unknown(), Joi.array()),
 	// parsing url as set of words and nums followed by dots, optional port and any possible path after
 	webServiceURL: Joi.string().regex(/https?:\/\/(?:[a-z0-9]+\.?)+(?::\d{2,})?(?:\/[\S]+)*/),


alexandercerutti:
Hi! Thank you for using it and thank you for your PR!
Let me explain you better why currently it is like this. When I designed this library, I chose among all the fields, the ones that should be editable and the one that probably shouldn't.
OrganizationName, in my opinion, should be strictly linked to the pass model. In fact I'm pretty difficult to be convinced to unlock one of these fields (last one was logoText, and I had to do that because of some English transportation rules ¯\_(ツ)_/¯  - but maybe I should allow all the fields and don't care about the usage?).

Anyway, meanwhile, if you need to create a pass with two different organization names, you'd probably need two-almost identical models.
Actually you could have one model and add a different pass.json model when you are creating the pass, in the object at "additionalBuffers". 🤔 What about this?

I'm suggesting you this because, I'd have to do right now a release for you to have this feature.

Let me know 😊

WaveMeUp:
Thank you for quick response!
I completely understand your point of view and your arguments are pretty valid.

Let me explain why we have to (in our case) override this property: we're creating web service for different companies, organizations, retails, etc., where they can publish their own electronic cards for Apple Wallet, Google Pay and so on. And because 'organizationName' is a visible field for users we just can't have it the same for every company.

Your solution with pass.json in 'additionalBuffers' is one of the options, sure, but these files will have a lot common fields anyway, so I'm not sure about this approach.

I doubled checked Apple documentation and didn't find any mentions about possible collisions or some bad scenarios with 'organizationName'.

Let me know what you think! :)

alexandercerutti:
No, in fact, Apple does not mention anything about possible collisions because there are none possible. That was just a design decision. 😜

I see your point: the company name is dynamic so you should generate first the pass.json, isn't it?

Okay, I think I can accept it.

But please, answer one moment to this question: should I allow (for you) all the fields? I mean, should I care about them? Just have an opinion and to understand I have to edit more after accepting this or not.

WaveMeUp:
Well, I think you should provide an option for users to edit all fields, because who knows what data you'll need to edit during run-time.
But it's up to you! 😄

alexandercerutti:
I'll accept it tomorrow and I'll review the fields that are missing 😊

alexandercerutti:
Released with other fields in the latest 2.0.3 🎉
Thank you very much again @WaveMeUp!

PR #48, opened 2019-12-10
update beacon and location methods to use overload

### LINKED ISSUE #47: index.d.ts:67 and 74 have bad parameter types
line 67: `beacons(...data: Schema.Beacon[] | null): this`
line 74: `locations(...data: Schema.Location[] | null): this`

TSC complains, rightfully, that a rest parameter must be of array type. This seems fixable just by removing the `| null` part.

the commit I'm referencing to is 681d9d144d9246440529b113e4e15a869a2df32d.

BTW, thanks for writing this library :]

nuts-n-bits:
Signed-off-by: nuts-n-bits

Closes #47

### DIFF
--- a/src/pass.ts
+++ b/src/pass.ts
@@ -287,13 +287,14 @@ export class Pass {
 
 	beacons(resetFlag: null): this;
 	beacons(...data: schema.Beacon[]): this 
-	beacons(...args: (schema.Beacon|null)[]){
-		if (args[0] === null) {
+	beacons(...data: (schema.Beacon|null)[]){
+
+		if (data[0] === null) {
 			delete this[passProps]["beacons"];
 			return this;
 		}
 
-		const valid = processRelevancySet("beacons", args);
+		const valid = processRelevancySet("beacons", data as schema.Beacon[]);
 
 		if (valid.length) {
 			this[passProps]["beacons"] = valid;
@@ -310,13 +311,13 @@ export class Pass {
 
 	locations(resetFlag : null) : this;
 	locations(...data: schema.Location[]): this;
-	locations(...args: (schema.Location|null)[]) : this {
-		if (args[0] === null) {
+	locations(...data: (schema.Location|null)[]) : this {
+		if (data[0] === null) {
 			delete this[passProps]["locations"];
 			return this;
 		}
 
-		const valid = processRelevancySet("locations", args);
+		const valid = processRelevancySet("locations", data as schema.Location[]);
 
 		if (valid.length) {
 			this[passProps]["locations"] = valid;
--- a/src/pass.ts
+++ b/src/pass.ts
@@ -281,17 +281,19 @@ export class Pass {
 
 	/**
 	 * Sets current pass' relevancy through beacons
-	 * @param data
+	 * @param data varargs with type schema.Beacon, or single arg null
 	 * @returns {Pass}
 	 */
 
-	beacons(...data: schema.Beacon[] | null): this {
-		if (data === null) {
+	beacons(resetFlag: null): this;
+	beacons(...data: schema.Beacon[]): this 
+	beacons(...args: (schema.Beacon|null)[]){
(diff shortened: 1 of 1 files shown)


nuts-n-bits:
So, the first two lines of the overload is what matters to the type checker, and the implementation line doesn't matter as long as it is inclusive of the union type of all type signatures.

In other words, the first two lines are sufficient to make sure that (only one null) or (Beacon[]) is valid. If you try to do `beacon(null, null)` as the code is right now, you get a parameter type check error, if you have strict type checking on. (I tried).

alexandercerutti:
Mmm, I made some tests too and you are right.

Using only

```typescript
locations(...data: (schema.Locations|null)[]) { /* implementation */ }
```

would mean these cases to success/fail:

```typescript
pass.locations(null); ✔
pass.locations({ latitude: 1, longitude: 2 }, null); ✔
pass.locations([null]); ❌
```

But we don't want the second to pass.

Yes, my method would require more overhead. So I think you convinced me. Just a question: why not calling all "data"? I mean, we could write it in the comments.
Or, at least, don't call it `args` eh eh 😂 I'm okay with "resetFlag".

Moreover, anyway, these changes must be applied to `index.d.ts` too (I created it manually because I don't like how typescript creates it).

Anyway, following the strict rules, I have to cast `data`  in `processRelevancySet` to remove the 'null'.

nuts-n-bits:
**Parameter naming**: When a rest parameter can be either one of two things (like data or resetFlag), I usually name it dataOrResetFlag for absolute clarity. I actually preferred that, so we can go with that if you don't mind. 

**Casting `data`**: Yeah, that you need to do. It's a slight pain in TS that in overload functions the types will be less clear requiring a manual cast.

**Generating .d.ts**: In tsconfig.json, in compilerOptions, add:
```
"declaration": true
```
will instruct TSC to emit .d.ts for you automatically.

BTW I didn't know you were handwriting .d.ts XD, thanks for the effort!

alexandercerutti:
About generating `.d.ts`, I know `declaration`, but as I said above there's often something I don't like in the auto-generated declaration. So sometimes I make it generate and merge mine and its 😂.

About parameter naming, "resetFlag" is fine for me: just rename `args` in `data`.

About casting, do you want to do that or I'll make it later, once merged? I could contribute to your fork if you would allow me 😊

nuts-n-bits:
I just pushed a second commit including parameter renaming and casting!

src/pass.ts
```diff
@@ -287,13 +287,14 @@ export class Pass {
 
 	beacons(resetFlag: null): this;
 	beacons(...data: schema.Beacon[]): this 
-	beacons(...args: (schema.Beacon|null)[]){
-		if (args[0] === null) {
+	beacons(...data: (schema.Beacon|null)[]){
+
```
alexandercerutti: Just to make it perfect, what about this blank line? 😂

nuts-n-bits: removed

alexandercerutti:
Okay, so I'll do the part of the declaration file, okay? 😄
I'll approve this PR tomorrow, so I can work on it and include other changes I made 😉

Thank you very much for your contribution 😄

nuts-n-bits:
That's great! Thank you for this library, saved me a lot of time. (I was about to use bash script openssl for my project without a native solution)

It's nice to work with you!

alexandercerutti:
Oh, I had some problems yesterday, so I didn't merge this yet (but brought on the changes).
I was remembering that I used the signature I proposed the first time somewhere else in my code: `barcodes` method.

I think I will change that too to this new signature. 😄

alexandercerutti:
Changes released in v2.0.4 🎉🎉 Thank you very much for your contribution ❤

P.s. @nuts-n-bits if you look at `index.d.ts`, I've used `declaration: true` and merged. It outputted two signatures instead of three. Just to let you know. 😊

PR #52, opened 2020-05-16
Adding serverless lambda to example

### LINKED ISSUE #51: Use this server on lambda
Hi bumped into this repo today.. quickly setup and all working awesome stuff!.. Wondering what will it take to host this as a service on a AWS lambda? Where would you suggest will be agood place to start to look for? Thanks!

Dicondur:
Fixes #51


alexandercerutti:
Thank you for the PR.
I don't understand why you've copied the model as is in the new folder for serverless. Is it to use the example as is in a lambda?
If so, I don't think it is needed to do like `cd examples/passkitServerless` in the folder Readme.

Actually I don't know which approach to adopt in terms of example folder organization. I mean, it could be organized in subfolders, as you did, but also moving the other examples to another sibling folder, which will include the webserver.
In that case, the models would be shared between the examples and you wouldn't need to copy them in the package.

Anyway, I'd like to contribute to your PR. Is the "Allow edits from maintainers" checkbox checked? So I can clone the project and fix some small details. 😊

P.s. What's JSPM? Okay, it is a package manager, but is it really needed to exclude it with a .gitignore?

Dicondur:
Yes, the lambda is not aware of any folders above `examples/passkitServerless`.

Go ahead, feel free to change as required, I too felt there got to be some structure. Sametime dint want to change that was already there, so created `passkitServerless` folder. We are replacing `express` with API GW so dint need it.

Sure, "Allow edits" are enabled. :)

ps: Its just similar to node_modules for npm the defaults came with it dint bother me.

alexandercerutti:
Perfect. Anyway, since here the role is just to show how it can be done, I'm not quite sure if it is worth to duplicate things...

alexandercerutti:
I was looking at `package.json`, but I don't understand where `handler.js` file is. Did you forget to commit it or was it intentional? Right not there's not an example of code, but just configuration.

.gitignore
```diff
@@ -4,4 +4,3 @@ passModels/
 certificates/
 *.code-workspace
 .vscode/
-*.js
```
Dicondur, line 7: This made to ignore the handler.js check if ok to be removed during merge.

alexandercerutti: I'd write the code in Typescript. I ignored .js files because I didn't want to commit transpiled code. Let me do few changes 😊

alexandercerutti: Mmm, well, converting it to Typescript required me to install @types/node and to add a tsconfig.json.
That @types module says that `stream.toString`, does not accept parameters. So what I'm asking myself is:
- Are you sure this works?
- Do you think that there's a different typing package to use for serverless?

Dicondur: Not sure if I understood this correct

> I'd write the code in Typescript. I ignored .js files because I didn't want to commit transpiled code. Let me do few changes 😊

It that case dont change .gitignore and and just add the handler.js file alone. Are you able to do it?

alexandercerutti: I think I can, of course, but for the sake of consistency and to use the same language, I'd like to move it to Typescript instead of JS.

Dicondur: > Mmm, well, converting it to Typescript required me to install @types/node and to add a tsconfig.json.
> That @types module says that `stream.toString`, does not accept parameters. So what I'm asking myself is:
> 
> * Are you sure this works?
> * Do you think that there's a different typing package to use for serverless?

Are you trying to convert handler.js to typescript?

alexandercerutti: I'd like to, yes. But as you might know, it requires typings for the packages imported in the project.

Dicondur: Haa I see what you are doing... Do you see any benefits of using typescript for lambda? I mean it is possible, but ultimately lambda needs to be run in Node.js runtime.
No matter the file comes off from a transcomplier or not, the lambda still needs a .js handler.

As a user I dint have to run that extra step to convert the .ts to .js before using it.

I tried recreating your error. Looks like `@types/node`  `Stream.toString('base64')` does not have similar implementations. The error goes away if I use pure Node.js way `const { createPass } = require("passkit-generator");` to import the library.

Let me know if you find something I could learn! Thanks :)

alexandercerutti: Well, typescript phase is independent from where you have to run it. It matters only how you set it up. You'll always have a JS in output, but it helps to create a better code, even if it requires one more little step.

> I tried recreating your error. Looks like @types/node Stream.toString('base64') does not have similar implementations. The error goes away if I use pure Node.js way const { createPass } = require("passkit-generator"); to import the library.

I've looked at the documentation for Node.JS but I don't see anything about `Stream.toString(“base64”)`. I mean, is it possible for it to discard the parameter and converting automatically to a base64 string?

Of course, if you use only Javascript, it won't report you any error if you pass it a parameter that it won't use.

Are you able to find any documentation about this method's parameter?
The only way I found to convert a stream to a base64 is passing through Buffers. For this reason I asked you if this was working.

Can you try to generate a pass on a Lambda by removing "base64" parameter?

Dicondur: It is Buffer and not Stream, my bad.
[https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end](url)

Actually it is API Gateway which uses "Lambda Proxy" that needs the response in "base64", doesnt work without it!

So, how have you been converting to base64 so far? Need to convert pass object to base64. `const pass = await examplePass.generate()`

alexandercerutti: Okay, as you see Typescript is needed for something 😁
Can you provide me the documentation about this Lambda Proxy, please?
Anyway, to convert a stream in base64, we should pass through a buffer to collect all the chunks I think.

So, right now, doing `stream.toString("base64")`  is not working, am I right?

P.s. `generate` method is not asynchronous. It does not require "await". That was in version 1.x 😊

Dicondur: [https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html](url)

Dicondur: I still feel users shld have an option to choose over Node.js or TS to use :)

alexandercerutti: Thanks, I’ll give a look ASAP.
Using typescript or not, doesn’t change anything as I already told you 😊
This is just an example of how you can use this library, whether you are using TS or not.
It is not a template project that you'll clone to start using this library in Lambdas.
Since this project is fully written in Typescript, I'd like this example to be written too in Typescript to be coherent to the rest of the code. But in the end, it doesn't even matter, because Typescript get's converted to Javascript. Actually I don't know how many people are using my package in Typescript or Javascript. 😊

alexandercerutti: Anyway, looking at the link you gave me, I'm learning that AWS accepts also binary data.
It seems, anyway, that we cannot escape from converting the stream to a buffer or, anyway, in a binary data. But we could make the pass being sent as `application/vnd.apple.pkpass`,  which is the official mime-type for Apple Wallet Passes.

Looking at [this example repository](https://github.com/activescott/apigateway-lambda-binary-response), we might do these things:

- We could setup (and write it in the README) the parameter `binaryMediaTypes` in the API (I don't know how to do that).
- We could then set up the response to send a header `Content-Type: application/vnd.apple.pkpass`.
- Convert the stream to buffer and send it as is, without converting it to base64.

What do you think about it?

Dicondur: > * We could setup (and write it in the README) the parameter `binaryMediaTypes` in the API (I don't know how to do that).

[https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-configure-with-console.html](url)
[https://stackoverflow.com/questions/45982129/wallet-pass-auto-update-web-service-using-aws-api-gateway](url)
[https://www.serverless.com/framework/docs/providers/aws/events/apigateway/](url) search for binary media in this doc.

Some more docs :) 
Worked with Binary Media Types: */* on mine. Havnt tried serverless way to setting this yet.

examples/serverless/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+  "compilerOptions": {
+    "target": "es2017",
+    "module": "commonjs",
+    // "outDir": "./dist",
+    "strict": true,
+    "baseUrl": "./",
+    "typeRoots": [
+      "node_modules/@types"
+    ],
+    "types": [
```
Dicondur, line 11: Did you try this?

examples/serverless/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+      "node"
```
Dicondur, line 12: Do you try this?

.gitignore
alexandercerutti: Okay, so if I understood correctly, you have to set `*/*` in Binary Media Types. But if you set it but send `application/vnd.apple.pkpass`, it won't work, did I get it right?

Please note that the stack overflow thread is about the *service* called PassKit, which has its own endpoints. It says that PassKit endpoint return as Accept Header `*/*`, which for AWS does not mean *all* the mime types but only itself.

So, my question is: what if you set in binary media types `application/vnd.apple.pkpass` and respond to lambda execution with the same Mime Type?

A Response like this I think:

```javascript
  return {
    statusCode: 200,
    headers: {
      "Content-Type": "application/vnd.apple.pkpass"
    },
    isBase64Encoded: encoding === "base64",
    body: buffer.toString(encoding)
  }
```

And you **call** the lambda URL with the header `Accept: application/vnd.apple.pkpass` ?
Let me know, I'm curious about this. Thank you

examples/serverless/tsconfig.json
alexandercerutti: Not needed. I'm replacing them with `moduleResolution`. ✌

alexandercerutti:
So, I've changed a few things, like the name of function (to avoid naming conflict with `createPass` internal function) and corrected types. I've also added path resolving (it is a good practice) and added aws-lambda typings to not put all the variables to any manually but to give a type-signature to which the function should conform to.

Yet we have to convert the stream to buffer.

alexandercerutti:
I've applied changed to READMEs. There are still a lot of details to be filled in the serverless's one.
I left some placeholder notes. Step-by-step we'll fill them. 😊

.gitignore
Dicondur: Sorry, I shld have added some context earlier, the reason to deploy on lambda + Api Gateway is so that we can get a url endpoint and use that url in the webservice url in pass.json template to update the pass and send push notifications.

PassKit Ref from **Apple** below: I know there is another service by same name arrgh!
[https://developer.apple.com/library/archive/documentation/PassKit/Reference/PassKit_WebService/WebService.html](url)

As per this doc:
[https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/PassKit_PG/Updating.html#//apple_ref/doc/uid/TP40012195-CH5-SW1](url)

Now to your question:
Changing from `*/*` or `application/vnd.apple.pkpass` set as Binary Media Type in API Gateway or changing the headers doesnt change the way the pass gets downloaded during the GET Call. Did I address you query?

Dicondur:
I have given all details required to get this setup running on lambda..Feel free to add any details as required.

.gitignore
alexandercerutti: I understand your point. If the purpose is this, I don't think the code is fully correct. I mean, when you set `webService` url in `pass.json`, it will be used in a lot of ways.

A lambda can be used correctly, as per the example we are building, to generate a pass and return it IMHO. But it should already have a URL to use. Maybe of another lambda? I don't know, it depends of what you want to achieve.

In the example you built in Javascript, only the pass is being generated, but you are not achieving what you say you want to achieve.

alexandercerutti:
@Dicondur Why closing? 😄 How am I supposed to ship the changes if you close the PR?
Anyway, okay. I'll do some searches and improve.
So, can you confirm that the example, through the conversion of the stream in buffer, it works correctly?

Anyway, if you want to give up, I'll change the direction branch and merge this to keep it and merge it as soon it is ready. So you can delete your branch if you want. Just tell me what you'd like to do.

.gitignore
Dicondur: Yes, this one lambda is only going to be used for creating the pass, I know. I think it will be better to setup on your enironment and check trying it. Much eaiser than lengthy messages. Dont you think? I have mine setup and working I shared my setup and details. I can lookup at some things if you want.

Dicondur:
I dint delete my fork! I dont have much to add to this example besides the code I shared. I have mine setup and working shared what is working and also gave example code. 

Let me know if I can get you any thing specific you want to check on the intial working code!

alexandercerutti:
Okay, I'm sorry. Please understand my point of view: I never used lambdas (I know less than you about them), and we are working on this. But after that message and PR closing I didn't know what to expect.

Anyway, I'll try to continue this example when I'll have a bit of time. I'm bringing on another big and great project about Passkit passes creation that is taking me all the time I do not use to work. 😄

Just let me know: how did you proceeded with the fact of streams? Did you proceed to convert them to buffer and return it as base64 or the current code still works? Just to know what I have to expect before setting up my own environment / lambda. Thank you.

Dicondur:
Me too learning, there is always some to learn for everyone.  :) 
Iam just sharing what worked for me so other can find it easier.

To host this service on lambda these were some main pitfalls to watch for:
1) I imported js version  const { createPass } = require("passkit-generator"); of the lib.
2) returned body after converting to base64. 

stream = examplePass.generate();
body: stream.toString('base64'); //Dont know how to do this in TS. TS doesnt compile here.

Returning buffer could likely be a work around.
 
3) In API Gateway Settings > Change Binary Media Types to */*
The downloaded file shows up properly with this option only. 
Tried other mime types doesnt work.

Hope this helps?
Thanks
Dinesh

Dicondur:
Some thing related to this I believe!

In Node.js, resolved the warning about using 'Buffer' (as it was deprecated) by replacing it with '.from' method. Was able to get rid of the warning.

Eg: To extract body which is in 'base64'. 

  var bodyBuff = new Buffer.from(event.body, 'base64');
  var bodyAscii = bodyBuff.toString('ascii'); // Only string escaped with '\'
  var bodyJSON = JSON.parse(bodyAscii);

alexandercerutti:
Isn't `new Buffer` the one that got deprecated?
Anyway, which was the warning? Deprecation warning?

And, so where would you place this code? At the end of the lambda?

Dicondur:
Ya deprecation warning. Using new Buffer.From() instead of new Buffer() 

Lambda passes an event object and we can extract the some of them like.
 const { path, pathParameters, queryStringParameters, headers, body } = event; 

Now the body can be in base64 as we have */* in API settings>Binary Type 

To your question, not required to be added to lambda.
Only required on cases if you do want to pass json body to your lambda.

alexandercerutti:
Actually I don't know which version of Node.JS you are working on, but please notice that [Buffer constructor has been deprecated since Node 6](https://nodejs.org/api/buffer.html#buffer_new_buffer_array). As you can see from the history of the constructor, deprecation notice for **this** method has been added or removed few times. In Node 7, the deprecation warning disappeared and got added back in Node 10 only for non-`node_modules/` usages.

You should use just `Buffer.from`, not `new Buffer.from`.

Deprecation on the side, so if you send to lambdas a JSON body, inside of them you do not receive a JSON but something else? That sounds weird to me 🤔

alexandercerutti:
A Serverless/AWS example is being added with the new v3.0. Thank you anyway! Closing this :)

PR #69, opened 2021-02-26
added Buffer Model to README

### LINKED ISSUE #68: Add Buffer Model to README
Love this project! Super useful in the professional setting I'm working in :) The only thing that would have been helpful would be to have an example of the Buffer Model in the README or even the API docs as it took me a bit to find it in the Migration Guide! Again, thank you for all your work on this!

sarawarnock:
Took the buffer model usage example from the Migration Docs and added to README - it took me a bit to find the Migration Guide and what I needed was an example of the Buffer Model!! 

Closes #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,8 @@ try {
 		},
 
 		// The rest will be the same as above 
+		...	
+	});
 
 } catch (err) {
 	doSomethingWithTheError(err);
--- a/README.md
+++ b/README.md
@@ -140,7 +140,9 @@ ___
 
 <a name="usage_example"></a>
 
-## Usage Example - Folder Model
+## Usage Example
+
+#### Folder Model
 
 ```typescript
 /**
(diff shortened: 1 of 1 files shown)


alexandercerutti:
You surprised me with such speed! 😄
Thank you for your Pull Request.

What about having just an header ("Usage Examples") with smaller headings?

Something like this:
___

## Usage Examples

#### Folder Model

```typescript
// Here we show the code
```

____

#### Buffer Model

```typescript
// Here we show the code
```

sarawarnock:
updated!! See what you think!

alexandercerutti (changes requested):
Yeah, I like it!
May I ask you to replace this:

```
			"it.lproj/pass.strings": Buffer.from([ ... ])
		},
		// The rest will be the same as above 			
```

with

```
			"it.lproj/pass.strings": Buffer.from([ ... ])
		},
		// The rest will be the same as above 
		...	
	});
```

So we have the correct syntax? 😄 Unpaired brackets drive me crazy ahah
Thank you!

sarawarnock:
Agreed! Fixed!

alexandercerutti (approved):

PR #72, opened 2021-05-30
fixed typos

KristofKekesi:


alexandercerutti:
Well, I must admit that I do not appreciate just typos pull requests because they give the "contributor" tag without almost no effort and reason to me...

So, if you can find other things that can be fixed, I might think to approve this, @KristofKekesi

Anyway, thank you

KristofKekesi:
This is all I found.

PR #86, opened 2021-11-01
Update code samples in README.md

### LINKED ISSUE #85: README.md file is inconsistent with 3.x
I ran into some issues following the examples in the README.md file leading me to think that there were problems with either my cert or key. Specifically:

```diff
const pass = PKPass.from({
	model: "./passModels/myFirstModel",
	certificates: {
-		wwdr: "./certs/wwdr.pem",
-		signerCert: "./certs/signercert.pem",
-		signerKey: "./certs/signerkey.pem",
+		wwdr: fs.readFileSync("./certs/wwdr.pem"),
+		signerCert: fs.readFileSync("./certs/signercert.pem"),
+		signerKey: fs.readFileSync("./certs/signerkey.pem"),
		signerKeyPassphrase: "123456"
	},
}, {
	// keys to be added or overridden
	serialNumber: "AAGH44625236dddaffbda"
});
```

Another minor detail is that `PKPass.from` returns a `Promise` so the code needs a little tweak to include `await`.

socsieng:
Fixes #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,8 @@ This is a standard procedure: you would have to do it also without using this li
 #### Folder Model
 
 ```typescript
+import * as fs from "fs";
+
 /**
  * Use `const { PKPass } = require("passkit-generator");`
  * for usage in pure Node.js
@@ -121,12 +123,12 @@ This is a standard procedure: you would have to do it also without using this li
 import { PKPass } from "passkit-generator";
 
 try {
-	const pass = PKPass.from({
+	const pass = await PKPass.from({
 		model: "./passModels/myFirstModel",
 		certificates: {
-			wwdr: "./certs/wwdr.pem",
-			signerCert: "./certs/signercert.pem",
-			signerKey: "./certs/signerkey.pem",
+			wwdr: fs.readFileSync("./certs/wwdr.pem"),
+			signerCert: fs.readFileSync("./certs/signercert.pem"),
+			signerKey: fs.readFileSync("./certs/signerkey.pem"),
 			signerKeyPassphrase: "123456"
 		},
 	}, {
@@ -168,9 +170,9 @@ try {
 		"it.lproj/pass.strings": Buffer.from([ ... ])
 	},
 	{
-		wwdr: "./certs/wwdr.pem",
-		signerCert: "./certs/signercert.pem",
-		signerKey: "./certs/signerkey.pem",
+		wwdr: fs.readFileSync("./certs/wwdr.pem"),
+		signerCert: fs.readFileSync("./certs/signercert.pem"),
+		signerKey: fs.readFileSync("./certs/signerkey.pem"),
 		signerKeyPassphrase: "123456",
 	},
 	{
--- a/README.md
+++ b/README.md
@@ -123,13 +123,14 @@ import * as fs from "fs";
 import { PKPass } from "passkit-generator";
 
 try {
+	const { wwdr, signerCert, signerKey, signerKeyPassphrase } = getCertificateData();
 	const pass = await PKPass.from({
 		model: "./passModels/myFirstModel",
 		certificates: {
-			wwdr: fs.readFileSync("./certs/wwdr.pem"),
-			signerCert: fs.readFileSync("./certs/signercert.pem"),
-			signerKey: fs.readFileSync("./certs/signerkey.pem"),
-			signerKeyPassphrase: "123456"
+			wwdr,
+			signerCert,
+			signerKey,
+			signerKeyPassphrase,
 		},
 	}, {
 		// keys to be added or overridden
@@ -151,6 +152,16 @@ try {
 } catch (err) {
 	doSomethingWithTheError(err);
 }
+
+// Certificate data is typically retrieved from your secrets management solution
+function getCertificateData() {
+	return {
+		wwdr: fs.readFileSync("./certs/wwdr.pem"),
+		signerCert: fs.readFileSync("./certs/signercert.pem"),
+		signerKey: fs.readFileSync("./certs/signerkey.pem"),
+		signerKeyPassphrase: "123456",
+	};
+}
 ```
 
 #### Buffer Model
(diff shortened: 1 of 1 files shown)


alexandercerutti:
Thank you for this PR! I'd rather not import `fs` in the examples or other libraries, because certificates might come from different sources, e.g. a secret manager somewhere.

What about using a more generic source like:

```ts
const { signerCert, signerKey, wwdr, signerKeyPassphrase } = getCertificatesDataSomehow();
```

?

Let me know!

README.md
```diff
@@ -114,19 +114,21 @@ This is a standard procedure: you would have to do it also without using this li
...
+
 /**
  * Use `const { PKPass } = require("passkit-generator");`
  * for usage in pure Node.js
  */
 import { PKPass } from "passkit-generator";
 
 try {
-	const pass = PKPass.from({
+	const pass = await PKPass.from({
```
socsieng, line 126: I know `await` isn't valid outside of an `async` function but I thought it was clearer than the more _correct_ alternatives:

1. Wrap the code sample in an `async` function,
2. Move the subsequent code into `PKPass.from().then(pass => { ... })`

Happy to make either change if you think it is appropriate.

socsieng:
> What about using a more generic source like:
> 
> ```ts
> const { signerCert, signerKey, wwdr, signerKeyPassphrase } = getCertificatesDataSomehow();
> ```
> 
> ?
> 
> Let me know!

To be honest. I would prefer a concrete example. `getCertificatesDataSomehow()` is still a little ambiguous if someone was familiar with the old/previous API.

For me, the `fs` example communicates to me that it is the contents of the file as opposed to the file path of the previous version of the API. I think including a comment to say that it should be retrieved from external sources like secrets manager would be enough.

README.md
alexandercerutti: I'm okay with `await` like this. Now Node.js has top-level await, isn't it? Also, the previous example was like this.

alexandercerutti:
What if, instead of `somehow`, we explicit `getCertificatesDataFromFS()` instead? Is it clearer?

socsieng:
I updated the PR. let me know what you think.

alexandercerutti:
Sorry, but I don't like to include that `import fs`. 😔 For me it kinda makes confusing that certificates can come only from FS. I mean, yeah, I take them from FS in examples in the examples folder, but they are there for this reason. I'd rather make Readme examples as generic as possible. Readme examples are not meant to have a context. If you see, I already make use of something like `doSomethingWithTheBuffer(buffer);`. That's the reason 😄

socsieng:
All good. In this case, I'll withdraw this PR.

alexandercerutti:
Thanks anyway, @socsieng 😄

socsieng:
This is great work by the way. I found the official documentation to be inadequate so this library, its documentation, and code samples are a huge help.

alexandercerutti:
... have you seen the archived one by Apple? 😄 The new documentation is way better, but honestly a bit confusing to be navigated.

Thank you for using it ❤
I'm open to other suggestions if you have any

socsieng:
> ... have you seen the archived one by Apple? 😄 The new documentation is way better, but honestly a bit confusing to be navigated.

I haven't. I feel like the documentation could benefit from code samples. Not just what the APIs are, but how to use them.

> Thank you for using it ❤
> I'm open to other suggestions if you have any

One thing that seems unnecessary is the implicit requirement around the `.pass` suffix for the model folder:

```
const pass = await PKPass.from({
	model: "./passModels/myFirstModel",
	certificates: {
		wwdr,
		signerCert,
		signerKey,
		signerKeyPassphrase
	},
}, { ... });
```

It isn't a big deal, I just ended up renaming my folder but, it isn't immediately obvious that the folder needs to include the `.pass` suffix. If the example included `./passModels/myFirstModel.pass`, I don't think I would have made this mistake in the first place.

The biggest win for me was the instructions around [certificates](https://github.com/alexandercerutti/passkit-generator/wiki/Generating-Certificates#generate-certificates-through-terminal) and CLI commands on converting the `.cer` to `.pem`. (FYI, Step 9 - converting `.key` to `.pem` is unnecessary when using `openssl genrsa -out <your-key-name>.key 2048`)

I'm just getting started but will let you know if anything else comes up.

alexandercerutti:
Mmm, I've written about .pass extension enforcement earlier in the Readme. But yeah, I can add the extension along with a comment, why not!

About the certificates, I'm glad it is useful. I had some issues while attempting to generate the certificates the first time, so I wanted to share my knowledge with others! For what concerns .key to .pem, do you mean that a key can still be used as-is? Nice to know, I'll do some tests in the next few days, thank you! 😄

socsieng:
> For what concerns .key to .pem, do you mean that a key can still be used as-is? Nice to know, I'll do some tests in the next few days, thank you! 😄

The key can be used as-is.

Running `openssl rsa -in <your-key-name>.key -outform PEM -out signerKey.pem` effectively produces an identical file if you've used `openssl genrsa -out <your-key-name>.key 2048` to generate the key.

alexandercerutti:
I proceeded in testing this and I improved the guide as you suggested, thank you! 😄

socsieng:
One of the things that I found useful with the original example was knowing which file types are expected.

I think adding comments would be useful. Example:

```js
const pass = await PKPass.from({
	model: "./passModels/myFirstModel",
	certificates: {
		wwdr,                 // wwdr.pem             from step 9: https://github.com/alexandercerutti/passkit-generator/wiki/Generating-Certificates#generate-certificates-through-terminal
		signerCert,           // signerCert.pem       from step 8: https://github.com/alexandercerutti/passkit-generator/wiki/Generating-Certificates#generate-certificates-through-terminal
		signerKey,            // <your-key-name>.key  from step 6: https://github.com/alexandercerutti/passkit-generator/wiki/Generating-Certificates#generate-certificates-through-terminal
		signerKeyPassphrase
	},
}, {
	// keys to be added or overridden
	serialNumber: "AAGH44625236dddaffbda"
});
```

alexandercerutti:
Developers really hate so much reading things, uh? I know, I'm one of them, but I also know the importance of reading what has been written. Readme has been studied and composed to be read sequentially, with all the information needed. Skipping things won't help. Adding those comments is also risky in terms of example flow: space and formatting will probably make those lines wrap down... and I honestly think that a broken code is more difficult to read.

Also, the two guides for Certificates generation do have not the same amount of steps: the first one has 8, the second has 9.

PR #100, opened 2022-01-31
Check for Joi.binary

billytrend:
Per this release https://github.com/sideway/joi/issues/2037#:~:text=The%20browser%20build%20does%20not%20include%20TLD%20email%20validation%2C%20Joi.binary()%2C%20the%20describe/build%20functionality%2C%20or%20debug/tracing. Joi.binary doesn't work in the browser and yield an error as follows:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,7 @@
 		"build:src": "rimraf lib && npx tsc -p tsconfig.dist.json",
 		"build:spec": "rimraf \"./spec/*.!(ts)\" && npx tsc -p tsconfig.spec.json",
 		"prepublishOnly": "npm run build",
-		"test": "npm run build:spec && npx jasmine",
-		"prepare": "npm run build"
+		"test": "npm run build:spec && npx jasmine"
 	},
 	"author": "Alexander Patrick Cerutti",
 	"license": "MIT",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
 		"build:src": "rimraf lib && npx tsc -p tsconfig.dist.json",
 		"build:spec": "rimraf \"./spec/*.!(ts)\" && npx tsc -p tsconfig.spec.json",
 		"prepublishOnly": "npm run build",
-		"test": "npm run build:spec && npx jasmine"
+		"test": "npm run build:spec && npx jasmine",
+		"prepare": "npm run build"
 	},
 	"author": "Alexander Patrick Cerutti",
 	"license": "MIT",
--- a/src/schemas/Certificates.ts
+++ b/src/schemas/Certificates.ts
@@ -8,11 +8,14 @@ export interface CertificatesSchema {
 	signerKeyPassphrase?: string;
 }
 
+// Joi.binary is not available in the browser
+const maybeBinaryOrString = Joi.binary ? [Joi.binary(), Joi.string()] : [Joi.string()];
+
 export const CertificatesSchema = Joi.object<CertificatesSchema>()
 	.keys({
-		wwdr: Joi.alternatives(Joi.binary(), Joi.string()).required(),
-		signerCert: Joi.alternatives(Joi.binary(), Joi.string()).required(),
-		signerKey: Joi.alternatives(Joi.binary(), Joi.string()).required(),
+		wwdr: Joi.alternatives(...maybeBinaryOrString).required(),
+		signerCert: Joi.alternatives(...maybeBinaryOrString).required(),
+		signerKey: Joi.alternatives(...maybeBinaryOrString).required(),
 		signerKeyPassphrase: Joi.string(),
 	})
 	.required();
--- a/src/schemas/Certificates.ts
+++ b/src/schemas/Certificates.ts
@@ -8,14 +8,14 @@ export interface CertificatesSchema {
 	signerKeyPassphrase?: string;
 }
 
-// Joi.binary is not available in the browser
-const maybeBinaryOrString = Joi.binary ? [Joi.binary(), Joi.string()] : [Joi.string()];
+// Joi.binary is not available in the browser so fallback to basic check
+const binary = Joi.binary ? Joi.binary() : Joi.custom((obj) => Buffer.isBuffer(obj));
 
 export const CertificatesSchema = Joi.object<CertificatesSchema>()
 	.keys({
-		wwdr: Joi.alternatives(...maybeBinaryOrString).required(),
-		signerCert: Joi.alternatives(...maybeBinaryOrString).required(),
-		signerKey: Joi.alternatives(...maybeBinaryOrString).required(),
+		wwdr: Joi.alternatives(binary, Joi.string()).required(),
+		signerCert: Joi.alternatives(binary, Joi.string()).required(),
+		signerKey: Joi.alternatives(binary, Joi.string()).required(),
 		signerKeyPassphrase: Joi.string(),
 	})
 	.required();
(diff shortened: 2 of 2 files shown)


alexandercerutti:
Hi @billytrend, thank you for this PR!

🤔 I'm not quite sure, for two reasons:

1) we are running on Joi 17.4
2) passkit-generator is meant to be run on node.js runtime.

Are you perhaps trying to run it on a browser? You'll run into several other issues for sure by doing this.

alexandercerutti:
Great choice! I was completely ignoring the existence of Cloudflare Workers. It would be great to create an example in examples folder to check p.g. future changes against it (I also need to check if it can be run offline just like serverless and can host several endpoints). 

Also I'm now curious about them. I'll study them a bit and I will try to setup something working.

billytrend:
I shared the hello world here: https://github.com/billytrend/cloudflare-worker-passbook. I haven't had time to check the instructions yet though so may need to expand them out later this week. Of course the example doesn't yet work fully because of the Joi.binary issue as you can see here: https://github.com/billytrend/cloudflare-worker-passbook/runs/5011303804?check_suite_focus=true

billytrend:
Changed the logic a bit and confirmed this fixes the issue in cloudflare workers when using a [browser Buffer](https://github.com/feross/buffer) but let me know if you can see a better way to do this! Thanks

src/schemas/Certificates.ts
```diff
@@ -8,14 +8,14 @@ export interface CertificatesSchema {
 	signerKeyPassphrase?: string;
 }
 
-// Joi.binary is not available in the browser
-const maybeBinaryOrString = Joi.binary ? [Joi.binary(), Joi.string()] : [Joi.string()];
+// Joi.binary is not available in the browser so fallback to basic check
```
alexandercerutti, line 11: Since cloudflare workers are kind of a hybrid of browser and node.js environments, I think this comment might bring people confusion, since we don't actually support browsers. CW are "weird way" of doing serverless functions.

I'd write 

> Joi.binary is not available in browser-like environments (like Cloudflare workers) ...

What do you think about it? 😄

billytrend: Yeah sounds good!

alexandercerutti:
I'm setting up rn an example to run on Cloudflare Workers, so I can get better into this.

billytrend:
Sounds good, let me know if you have any issues. I updated the instructions on https://github.com/billytrend/cloudflare-worker-passbook hopefully it's a bit more clear!

alexandercerutti:
Your example is revealing as super useful! I think I'll create one example with wrangler integrated webpack and one with webpack 5, just for knowledge!

billytrend:
Cool, I temporarily switched my example repo to use this fork.. and it now [builds and deploys](https://github.com/billytrend/cloudflare-worker-passbook/actions/runs/1780695608) you can try it here https://cloudflare-worker-passbook.billytrend.workers.dev

billytrend:
By the way, it is not the most economical to run on cloudflare workers. I am seeing execution time of around 300ms which is more than cloudflare's free allowance. But still it's a convenient way to host the code and is also geographically distributed.. 
image

alexandercerutti:
As per Discussion #98, maybe Joi validation is not very suitable to these cloud contexts, IMHO. Maybe it should be used just for development deployment...

package.json
```diff
@@ -9,7 +9,8 @@
 		"build:src": "rimraf lib && npx tsc -p tsconfig.dist.json",
 		"build:spec": "rimraf \"./spec/*.!(ts)\" && npx tsc -p tsconfig.spec.json",
 		"prepublishOnly": "npm run build",
-		"test": "npm run build:spec && npx jasmine"
+		"test": "npm run build:spec && npx jasmine",
+		"prepare": "npm run build"
```
alexandercerutti, line 13: Not so clear why you integrated this script 🤔 @billytrend

billytrend: I'm not sure if I needed it or not but I was trying to make this work 
https://github.com/run-pass/run-pass.github.io/blob/4ea73b90c485b462e06a450b9aa582fdf7754087/cloudflare-worker/package.json#L27

alexandercerutti: Okay I understand why you needed it. [According to NPM](https://docs.npmjs.com/cli/v8/using-npm/scripts), `prepare` is run on these situations:

> - Runs any time before the package is packed, i.e. during npm publish and npm pack
> - Runs BEFORE the package is packed
> - Runs BEFORE the package is published
> - Runs on local npm install without any arguments
> - Run AFTER prepublish, but BEFORE prepublishOnly

And, since this is a Typescript project, the thing falls under the 4th case (or under the 3rd, if you start counting from 0 😄).

In the real world (not specific fixes), we don't need it to run build on all these cases. So you can remove it. 😉
You can create a different branch in which you include this fix **and** the compiled version, so the lib folder. Then, you can install the dependency from a specific branch. By doing this way, you keep the fix branch clear to be merged. 👍

alexandercerutti:
I know this is not strictly related to the PR... but were you able to generate a ticket with an asset like footer (boardingPass) or thumbnail (eventTicket) picture? I'm somehow unable to do that. I can generate a valid pass, and the images are successfully inserted inside the bundle. They have a size but they seem to be, idk, corrupted? I did apply your edit, but I don't think it is strictly related... at least, I hope that.

My best guess is that the conversion between stringified-asset-payload and Buffer gets broken somehow, but I'm not sure how this could be fixed. 🤔

package.json
billytrend: yeah exactly. Alright, i've reverted the change! ty

billytrend:
ohh I hadn't actually realised it wasn't working but thanks for suggesting the fix, I'll merge that into my repo when I get round to it!

billytrend:
how do you feel about merging this one @alexandercerutti ?

alexandercerutti:
@billytrend I'm going to merge this asap! I'm almost done with the Cloudflare worker example. I've found other issues that could prevent in using passkit-generator with webpack 5 and CW 😄

billytrend:
@alexandercerutti also you might find this interesting, I did some cost calculations: https://github.com/billytrend/cloudflare-worker-passbook/blob/main/README.md#how-much-does-it-cost-to-run-in-cloudflare-workers

alexandercerutti:
@billytrend I genuinely don't know how were you able to debug a worker like this. I was unable to debug them effectively.  

This is very interesting! Maybe Cloudflare Worker is not very suitable for this kind of operation (generating passes)? Anyway, I bought the paid plan for testing 😂 I also don't know if there's a way to improve this on the library side.

billytrend:
It's actually surprisingly easy https://blog.cloudflare.com/profiling-your-workers-with-wrangler/.

I did find that it worked ok without the paid plan for development but if you hit the endpoint too much, it starts erroring.

I think there's probably not a good way of making it faster because I guess a lot of the signing etc is CPU bound but I don't think it matters too much. $5 seems pretty cheap for 1.8million requests per month. Also if you have other workers on your account, they can use the same allowances.

alexandercerutti:
Hey there @billytrend, I've published examples on a branch: https://github.com/alexandercerutti/passkit-generator/tree/example/cloudflare_worker/examples/cloudflare-worker

Would be glad if you could give a look at them and give me feedback about both of them! 😄

alexandercerutti:
Released in v3.1.4 🎉
Thank you very much, @billytrend, for your great contribution!

PR #101, opened 2022-02-03
Missing extension in minimal example

majuss:
Without .png extension it wont work. The pass will be created but opening is failing, because the icon is saved inside the zip without any extension.


alexandercerutti:
Hey there @majuss, thanks for using Passkit-generator!

Yes, I actually find it out yesterday while doing some tests and I was planning to fix it with the other PR that has been opened, but you were quicker than me 😄

Thank you very much ❤️

alexandercerutti (approved):

PR #104, opened 2022-02-19
Fixed FieldsArray.push to preserve the order of input items

### LINKED ISSUE #103: Field order in the loaded pass.json is not preserved
In the generated passes, the fields (eg: `primaryFields`, `secondaryFields`) appear in reverse order compared to the original source in `pass.json`. For example, loading a store card bundle with

```json
...
      "storeCard": {
        "secondaryFields": [
          {
            "key": "name",
            "label": "Name",
            "value": "Jon Snow"
          },
          {
            "key": "dateOfBirth",
            "label": "Date of birth",
            "value": "13 OCT 1980"
          }
        ],
...
```

produces a pass like this

image

---

Here is a simple unit test to reproduce the issue under the `spec/FieldsArray.ts` setup (untested):

```typescript
                it("should preserve field order", () => {
                        expect(fa.push({ key: "t1", value: "v1" })).toBe(1);
                        expect(fa.push({ key: "t2", value: "v2" })).toBe(1);
                        expect(fa.length).toBe(2);
                        expect(fa[0]).toEqual({ key: "t1", value: "v1" });
                        expect(fa[1]).toEqual({ key: "t2", value: "v2" });
                });
```

---

I noticed that it's caused by the `FieldArray#registerWithValidation` is traversing the array items in reverse order.
https://github.com/alexandercerutti/passkit-generator/blob/ccd38262ec69d37b669774aba6812feeb0a16534/src/FieldsArray.ts#L70-L87

Is there any particular reason? I believe it's an unexpected breaking change introduced since v3.

Many thanks in advance!

zhenwenc:
PR to fix the reported issue: https://github.com/alexandercerutti/passkit-generator/issues/103

> In the generated passes, the fields (eg: primaryFields, secondaryFields) appear in reverse order compared to the original source in pass.json.

Fixes #103

### DIFF
--- a/spec/FieldsArray.ts
+++ b/spec/FieldsArray.ts
@@ -40,6 +40,15 @@ describe("FieldsArray", () => {
 			expect(fa[0]).toEqual({ key: "t1", value: "v1" });
 		});
 
+		it("should preserve order of input items when adding fields", () => {
+			expect(
+				fa.push({ key: "t1", value: "v1" }, { key: "t2", value: "v2" })
+			).toBe(2);
+			expect(fa.length).toBe(2);
+			expect(fa[0]).toEqual({ key: "t1", value: "v1" });
+			expect(fa[1]).toEqual({ key: "t2", value: "v2" });
+		});
+
 		it("should add the key to the pool", () => {
 			fa.push({ key: "t1", value: "v1" });
 
--- a/src/FieldsArray.ts
+++ b/src/FieldsArray.ts
@@ -67,7 +67,7 @@ function registerWithValidation(
 
 	let validItems: Schemas.Field[] = [];
 
-	for (let i = items.length, field: Schemas.Field; (field = items[--i]); ) {
+	for (const field of items) {
 		try {
 			Schemas.assertValidity(
 				Schemas.Field,


src/FieldsArray.ts
```diff
@@ -67,7 +67,7 @@ function registerWithValidation(
 
 	let validItems: Schemas.Field[] = [];
 
-	for (let i = items.length, field: Schemas.Field; (field = items[--i]); ) {
+	for (let i = 0, field: Schemas.Field; (field = items[i++]); ) {
```
alexandercerutti, line 70: @zhenwenc I actually now see an issue here that was happening also before: if an element of the array is undefined, all the next are ignored. Probably it is better to switch to classic incrementing. What do you think about it?

spec/FieldsArray.ts
```diff
@@ -40,6 +40,15 @@ describe("FieldsArray", () => {
 			expect(fa[0]).toEqual({ key: "t1", value: "v1" });
 		});
 
+		it("should preserve order of input items when adding fields", () => {
+			expect(
+				fa.push({ key: "t1", value: "v1" }, { key: "t2", value: "v2" })
```
alexandercerutti, line 45: I guess this is good, but that might be the case to extend the test also for splice... and maybe pop? I can do it

src/FieldsArray.ts
zhenwenc: Very good point. Wondering would you like to use the for-each-loop syntax? For example:

```
for (field in items) {
  if (!field) continue;

  ... rest ...
}
```

alexandercerutti: I guess that would be fine as well. We don't actually need to have an index. Also, the size of the target array likely won't ever be over 10 elements (to stay wide; actually passes allow about up to 5 elements per field but `backFields`), so the performance impact here would be very transparent and inconsiderable IMHO.

Anyway, using a `continue` condition is actually not needed here because Schema validation would directly throw and get caught. I mean, yes, a condition might make us skip the whole validation process, but the user wouldn't get notified with a warning. So, I'd not put that.

alexandercerutti (changes requested):

src/FieldsArray.ts
zhenwenc: Thanks for the details! 

> Anyway, using a continue condition is actually not needed here because Schema validation would directly throw and get caught.

It will only throw a validation error if we mark the `Schema.Field` to be required. 
https://github.com/sideway/joi/issues/1358

I've added a unit test to verify this behavior, the validation passes through and failed at:
```
Failures:
1) FieldsArray push should log a warning if input items contain undefined
  Message:
    Expected spy log to have been called with:
      [ 'Cannot add field. t1' ]
    but actual calls were:
      [ TypeError: Cannot read properties of undefined (reading 'key') ].

    Call 0:
      Expected $[0] = TypeError: Cannot read properties of undefined (reading 'key') to equal 'Cannot add field. t1'.
  Stack:
    Error: Expected spy log to have been called with:
      [ 'Cannot add field. t1' ]
    but actual calls were:
      [ TypeError: Cannot read properties of undefined (reading 'key') ].
    Call 0:
      Expected $[0] = TypeError: Cannot read properties of undefined (reading 'key') to equal 'Cannot add field. t1'.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/fredc/code/passkit-generator/spec/FieldsArray.js:52:34)
        at <Jasmine>
```

---

I have a side question in terms of validation. The lib currently only logs the validation errors instead of propagating to the upstream, this is quite difficult for programs that allow user-defined pass templates, where it cannot report the validation error back to the user (assuming from an API endpoint).

zhenwenc: > So you mean picking this and appending a .required() to it, is it correct?

Yes, correct. You can either append `.required()` to the schema definition in `src/schemas/Field.ts` or only within the `FieldArray` scope (not sure if there is any performance penalty)

```typescript
Schemas.assertValidity(
	Schemas.Field.required(),
	field,
	Messages.FIELDS.INVALID,
);
```

This is a unit test that I wrote to reproduce the issue:
```typescript
it("should log a warning if input items contain undefined", () => {
                console.warn = jasmine.createSpy("log");

                fa.push(undefined, { key: "t1", value: "v1" });

                expect(console.warn).toHaveBeenCalledWith(
                        "TypeError: ValidationError happened. " +
                        Messages.FIELDS.INVALID.replace("%s", "\"value\" is required"),
                );

                expect(fa.length).toBe(0);
});
```

---

> About your question (even if it does not contain any question mark, so I'm not quite sure what is the actual question), do you mean both blocking and non-blocking errors should be emitted to the containing application? 

Sorry for not making it clear, I was wondering about the intention of not throwing the validation errors.

As you can see in my sandbox project (for learning purposes), it allows users to provide the Apple Pass template bundle when invoking the pass generation endpoint. In case the user made any mistake in the given `pass.json`, the JSON schema validation error from `passkit-generator` will only be logged on the server-side and silently proceed with a broken state from the user's perspective. This also affects real-world applications as well, where we cannot rely on any automation test to capture potential errors.

---

Would you like to proceed with this PR with the current changes as it was only intended to fix the ordering issue of `FieldArray#push`? We could have a further discussion of the other issues in a separate ticket/discussion.

alexandercerutti: Starting from the bottom: okay, that's a good point and very precious feedback. If you could open a Discussion thread, we could discuss it there without going offtopic here. I guess that would be a matter of a possible v3.x.0 or v4.

About the PR: I'd continue like this. Just apply the for-in. I'll take care of checking the rest 😉.

zhenwenc: Thanks! I've amended the commit. 🙂

alexandercerutti: Thanks! I'm not any way a fan of force-pushes :D I could have squashed everything in a single commit with the merge ;)

zhenwenc: Ahh... they are pretty much the same, just different preferences 😄 you can [view the force-push diff](https://github.com/alexandercerutti/passkit-generator/compare/21eec509add8330094fad49b5a4be1436b0ba1f4..16177dfa49495be28019fc69882d56291075d9aa) too.

alexandercerutti (approved):

src/FieldsArray.ts
alexandercerutti: I was experimenting a bit with what we talked about here and it ended up with me not setting Field as required because any attempt of mine of excluding `undefined` by validation, causes a lot of other things to fail!

I guess I'll add an explicit check throw with `continue` as you suggested.

PR #106, opened 2022-03-08
making either value or attributed value required

r-carroll:
## Description

Some passes contain only an `attributedValue` in place of the `value` field on the `backFields` object, which the current build does not allow for as it always requires `value` to be present. This change makes either field required.

## Check relevant checkboxes

-   [X] I've run tests (through `npm test`) and they passed
-   [X] I generated a working Apple Wallet Pass after the change
-   [X] Provided examples keep working after the change

## Relevant information
Example error print when you leave off both fields

```
Cannot add pass.json to bundle because it is invalid. ValidationError: "generic.backFields[0]" must contain at least one of [value, attributedValue]
```

[JOI or documentation](https://joi.dev/api/?v=17.6.0#objectorpeers-options)

### DIFF
--- a/src/schemas/Field.ts
+++ b/src/schemas/Field.ts
@@ -71,4 +71,4 @@ export const Field = Joi.object<Field>().keys({
 			is: Joi.number(),
 			otherwise: Joi.string().forbidden(),
 		}),
-}).or('value', 'attributedValue');
+}).or("value", "attributedValue");
--- a/src/schemas/Field.ts
+++ b/src/schemas/Field.ts
@@ -45,7 +45,7 @@ export const Field = Joi.object<Field>().keys({
 		Joi.string().allow(""),
 		Joi.number(),
 		Joi.date().iso(),
-	).required(),
+	),
 	semantics: Semantics,
 	// date fields formatters, all optionals
 	dateStyle: Joi.string().regex(
@@ -71,4 +71,4 @@ export const Field = Joi.object<Field>().keys({
 			is: Joi.number(),
 			otherwise: Joi.string().forbidden(),
 		}),
-});
+}).or('value', 'attributedValue');


alexandercerutti:
Hey there, @r-carroll ! Thank you for your PR, very appreciated <3.

By reading [Apple Archived Documentation](https://developer.apple.com/library/archive/documentation/UserExperience/Reference/PassKit_Bundle/Chapters/FieldDictionary.html#//apple_ref/doc/uid/TP40012026-CH4-SW1), it is explicitly written that "Value is required".

But now that I read it better, it also says:
> This key’s value overrides the text specified by the value key.

So I guess that's the reason you were able to generate a valid Pass and that's what drove me to set value as required. I never generated a pass with a link in it. Good to know!

Just two things:
 - You were lucky that `.or` exists for a long time because we are stuck at v17.4.2 😛
 - Would you mind running prettier and changing single quotes to double quotes to keep a linear style, please? I guess I'll have to set up an Action to auto-format PRs 😂
 
Thank you very much 🔝

r-carroll:
Thanks for taking a look! All my passes have links on the back and I was scratching my head for a bit wonder why I couldn't generate them with your library (which is awesome btw).

Good callout on the style consistency!

alexandercerutti:
Thank you very much! I hope to have done a good job on this project. I'm always open to suggestions! 😄
I'll merge your PR asap (more likely tomorrow evening, GMT+1)

alexandercerutti:
Hey, @r-carroll
I'm very unsure about this PR now. I tried to generate passes with links in them and I looked for other information. I found two critical points in this:

1) Apple [actually says that](https://developer.apple.com/documentation/walletpasses/passfieldcontent)

> The attributed value isn’t used for watchOS; use the value field instead.

That's the reason, I guess, you should always use a `value` (and it is marked as required) and then, in case, add an `attributedValue` 🤔 .

I don't have an Apple Watch (I'm going to ask friends of mine to test this behavior for me), but I guess that if you add a Pass with a field that has an `attributedValue` but doesn't have a `value`, something bad is gonna happen when it gets open - if you can, of course - on an Apple Watch (I'll update this message when I'll have a few tests results).

My best guess is that either the whole pass is considered invalid on Watch or that field is not shown at all (see point 2 for more details).

2) I tried to generate a Pass with only `attributedValue` but not `value`. It seems like only `backFields` support `attributedValue` (I've found no information about this, but just found it out). Otherwise, the field is not displayed at all if this is placed in the other "areas" - so I guess this is also valid for Apple Watch.

What do you think about it?
Let me know!

r-carroll:
Hi there,

This is some great detailed analysis! In my experience yes, `attributedValue` is only valid on the back of the card. My thinking is that on a watch, you can't view the back of the card regardless, so there would be no impact. I believe I had a friend open one of my passes on his watch (that had links on the back) and it worked just fine. I can't say for certain though, so I'll need to grab him tomorrow and do some testing to verify. So in summary, what I'm thinking is, 

1. If the watch can view back fields, or having an `attributedValue` with a link and no corresponding `value` renders the pass invalid on a watch, we close this PR. Or
2. If having an `attributedValue` containing a link and no corresponding `value` field has no impact on the watch, then we proceed with this PR.

alexandercerutti:
I've looked into the archived documentation: [according to Apple](https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/PassKit_PG/Creating.html#:~:text=%20the%20user%20cannot%20access%20the%20back%20side%20of%20the%20pass.%20), back fields should not be reachable on Apple Watch.

For this reason, I guess the main issue applies only to passes opened on iPhone and Mac: my biggest fear about this is that having `value || attributedValue` could mislead others into thinking that you could *always* (on all the "areas" but `backFields`) set just one of the two, which is not true, because if you don't set `value`, the field gets omitted at all.

r-carroll:
Ah, that's a good point. Is there a syntax we could use to make it apply only to `backFields`?

alexandercerutti:
Hey @r-carroll, I had the chance to see by myself on a colleague's Apple Watch the result aaaaand... guess what? Yeah, backFields are visible! I guess that that documentation is definitely outdated... because this change happened with watchOS 7! This is the updated documentation: https://developer.apple.com/documentation/walletpasses/passfields/backfields

So the problem would happen everywhere.

I attach you a few screenshots:

photo_2022-03-11 20 35 11
photo_2022-03-11 20 35 05
photo_2022-03-11 20 35 08

I think this is enough to allow us to close this, don't you agree?
Thank you very much for (at least the attempt of) improving passkit-generator, I really appreciated it!
If you could leave a ⭐️ on the project, that would be very helpful!

PR #118, opened 2022-11-09
Fix to unsupported fields rows

### LINKED ISSUE #116: Support for "row" field on auxiliaryFields?
### Discussed in https://github.com/alexandercerutti/passkit-generator/discussions/115

<div type='discussions-op-text'>

<sup>Originally posted by **danawoodman** November  8, 2022</sup>
From what I can tell, passkit-generator should work with the `row` property ([See](https://developer.apple.com/forums/thread/108302)) but I cannot figure out how. If I do the following:

```ts
const pass = new PKPass(...)
pass.type = 'eventTicket'
pass.auxiliaryFields.push(
  { value: 'Hello', label: 'Foo', key: 'foo', row: 0 },
  { value: 'World', label: 'Bar', key: 'bar', row: 1 },
)
```

I get a TS error and a Joi validation error:

```
ValidationError happened. Cannot add field. "row" is not allowed
[Error [ValidationError]: "row" is not allowed] {
  _original: { key: 'foo', value: 'Hello', label: 'Foo', row: 0 },
  details: [
    {
      message: '"row" is not allowed',
      path: [Array],
      type: 'object.unknown',
      context: [Object]
    }
  ]
}
ValidationError happened. Cannot add field. "row" is not allowed
```

The type for `pass.auxillaryFields` is `Field` but in the typing I see it is `PassField`: https://github.com/alexandercerutti/passkit-generator/blob/267d8a6e8c52337f94b7ba75359f8d3cb10a26ae/src/schemas/PassFields.ts#L16 

How can I add a `row` attribute to my auxiliary fields?
</div>

alexandercerutti:
## Description

Fixes #115
Fixes #116

## Check relevant checkboxes

## Relevant information

@danawoodman I generated a valid pass with two rows. Let me know if this fixes correctly the issue! 👍

### DIFF
--- a/examples/models/examplePass.pass/pass.json
+++ b/examples/models/examplePass.pass/pass.json
@@ -41,6 +41,20 @@
 				"label": "LOCATION",
 				"value": "Moscone West"
 			}
+		],
+		"auxiliaryFields": [
+			{
+				"key": "Foo",
+				"value": "foo2",
+				"label": "Foo(l)",
+				"row": 0
+			},
+			{
+				"key": "Foo2",
+				"value": "foo3",
+				"label": "Foo(l)2",
+				"row": 1
+			}
 		]
 	}
 }
--- a/src/FieldsArray.ts
+++ b/src/FieldsArray.ts
@@ -10,6 +10,7 @@ import * as Messages from "./messages";
 
 const passInstanceSymbol = Symbol("passInstance");
 const sharedKeysPoolSymbol = Symbol("keysPool");
+const fieldSchemaSymbol = Symbol("fieldSchema");
 
 export default class FieldsArray extends Array<Schemas.Field> {
 	private [passInstanceSymbol]: InstanceType<typeof PKPass>;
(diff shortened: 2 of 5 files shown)


danawoodman:
Works great! This is exactly what we needed. Only issue I notice is I still get a TS error with `pass.auxiiaryFields.push({ ..., row: 0 })`

alexandercerutti:
Uhm, I think that we need to apply some conditional. Let me check if I can do something.

alexandercerutti:
I experimented a bit with types. I wanted to achieve to have a dynamic type on `auxiliaryFields` when the type was specific, but I wasn't able to get it completely. Also, it required me a lot of changes in terms of typings and among all of them, I'm not quite sure if I've introduced some breaking changes yet.

So I guess the best solution is to set `auxiliaryField` to be of type `FieldWithRow[]`, even if we are not on an `eventTicket`.

danawoodman:
Yeah, I'd say that is preferable to a type error for something valid. You can just strip/omit `row` for pass types that don't support it

alexandercerutti:
Yeah, I can do that on validation time. I think I'll do this ;)

danawoodman:
Nice, lemme know when a release is ready and ill have a look

alexandercerutti:
@danawoodman pushed some changes!

danawoodman:
Thanks! Will you be cutting a release for this?

alexandercerutti:
Yes, of course! I had a very short amount of time to dedicate to this. Let me check if there's something more I can do to improve the changes I made and I'll publish them asap.

alexandercerutti:
Done, @danawoodman! Published in 3.1.7! Thank you very much for spotting this issue ;)

danawoodman:
@alexandercerutti thanks for being so quick to add this feature! 🍻

Issue #135, opened 2023-03-24
Wrong type for wifiAccess?


buren:
## Running Node Version

v16.7.0

## Description

Should be able to pass an array, `[{ ssid: "...", password: "..." }]`, but I'm not allowed.

### Current "working" type for passkit-generator
<img width="355" alt="Screen Shot 2023-03-23 at 15 37 47" src="https://user-images.githubusercontent.com/922411/227541115-ca09a4d2-13f5-41cc-a68d-626e6e8ba054.png">

### What the type should be
At least as I can tell this should work, see documentation screenshot below
<img width="785" alt="Screen Shot 2023-03-23 at 15 37 37" src="https://user-images.githubusercontent.com/922411/227541119-dd89c2b2-620d-4d09-85bf-b018a1696705.png">

### Apple documentation
🔗 [Apple docs for semantic tags](https://developer.apple.com/documentation/walletpasses/semantictags)

![Screen Shot 2023-03-24 at 14 57 52](https://user-images.githubusercontent.com/922411/227541108-871dd43a-2d02-4be5-824c-e3bfd5ac4cb8.png)

## Expected behavior

Should be allowed to pass an array. 

### Current "working" type for passkit-generator
<img width="355" alt="Screen Shot 2023-03-23 at 15 37 47" src="https://user-images.githubusercontent.com/922411/227541115-ca09a4d2-13f5-41cc-a68d-626e6e8ba054.png">

### What the type should be
At least as I can tell this should work, see documentation screenshot below
<img width="785" alt="Screen Shot 2023-03-23 at 15 37 37" src="https://user-images.githubusercontent.com/922411/227541119-dd89c2b2-620d-4d09-85bf-b018a1696705.png">

## Other details

🔗 [Apple docs for semantic tags](https://developer.apple.com/documentation/walletpasses/semantictags)
🔗 [Wifi Network semantic tag](https://developer.apple.com/documentation/walletpasses/semantictagtype/wifinetwork)

Love the project ⭐ Thanks a lot! 👑

buren:
[...]

PR #136, opened 2023-03-24
Set Semantic.wifiAccess as array type

buren:
## Description

See #135 for background and details.

## Check relevant checkboxes

-   [x] I've run tests (through `npm test`) and they passed – ⚠️ Test failed even without my change. Not sure if it's a local problem see below for output
-   [x] I generated a working Apple Wallet Pass after the change - ⚠️ Manually inspecting the generated `pass.json` there isn't any `wifiAccess` information present. 
-   [ ] Provided examples keep working after the change - ℹ️ Not tested.
-   [x] This improvement is or might be a breaking change - ⚠️ Yes, since this changes public API

## Notes

In Apple's own documentation it's very unclear where you should add `wifiAccess`. It looks like it's "supported" on all `headerFields`, `primaryFields`, `auxiliaryFields` and `backFields` which seems a bit weird. @alexandercerutti do you know?

🔗 [Relevant Apple documentation](https://developer.apple.com/documentation/walletpasses/pass/supporting_semantic_tags_in_wallet_passes)

## Failed test

```
passkit-generator git:(wifi-access-array-type) npm test                                                                                                                   git:(wifi-access-array-type|)

> passkit-generator@3.1.7 test
> npm run build:spec && npx jasmine


> passkit-generator@3.1.7 build:spec
> rimraf "./spec/*.!(ts)" && npx tsc -p tsconfig.spec.json

spec/FieldsArray.ts:13:8 - error TS2555: Expected at least 3 arguments, but got 2.

 13   fa = new FieldsArray(
           ~~~~~~~~~~~~~~~~
 14    {
    ~~~~
...
 19    pool,
    ~~~~~~~~
 20   );
    ~~~

  lib/FieldsArray.d.ts:12:83
    12     constructor(passInstance: InstanceType<typeof PKPass>, keysPool: Set<string>, fieldSchema: typeof Schemas.Field | typeof Schemas.FieldWithRow, ...args: Schemas.Field[]);
                                                                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'fieldSchema' was not provided.


Found 1 error in spec/FieldsArray.ts:13
```

### DIFF
--- a/src/schemas/Semantics.ts
+++ b/src/schemas/Semantics.ts
@@ -176,7 +176,7 @@ export interface Semantics {
 	venuePhoneNumber?: string;
 	venueRoom?: string;
 
-	wifiAccess?: SemanticTagType.WifiNetwork;
+	wifiAccess?: SemanticTagType.WifiNetwork[];
 }
 
 export const Semantics = Joi.object<Semantics>().keys({


alexandercerutti:
There are a lot of non-said things in Apple Wallet world. So no, I don't know honestly.
The failed issue is not relevant. I'll give a look at it when I'll merge this.

Thank you for your report and contribution!

alexandercerutti (approved):
There are a lot of non-said things in Apple Wallet world. So no, I don't know honestly.
The failed issue is not relevant. I'll give a look at it when I'll merge this.

LGTM.
Thank you for your report and contribution! I'll merge this ASAP.

alexandercerutti:
@buren I was reading the other checkboxes:

> I generated a working Apple Wallet Pass after the change - ⚠️ Manually inspecting the generated pass.json there isn't any wifiAccess information present.

That's kind of weird. Joi might be stripping it out... I don't know if this is okay. Do you have a repro case I can test with?

> This improvement is or might be a breaking change - ⚠️ Yes, since this changes public API

No this isn't. It's just a typings issue, but without Javascript, Joi validation should be already accepting an array as you pointed out.

buren:
I tried to add other data under the `semantics` key to an `secondaryFields` entry (see below), but it doesn't add that either to `pass.json` 🤔 Can I add `semantics` to any fields? (Apple's docs certainly makes it seem so.)

__Relevant I'm testing with__
```ts
pass.secondaryFields.push({
    key: 'event',
    label: eventLabel,
    value: event.name,
    semantics: {
      eventName: '',
      venueName: '123',
      wifiAccess: [
        {
          ssid: 'anyname',
          password: 'somesecret',
        },
      ],
    },
  });
```

At this time I don't have a public repo to share I'm afraid.

buren:
Ok, checking again it's possible to add `semantics` to the root of the pass and there the type is also correct (`wifiAccess` is allowed to be an array), so the "problem" seems to only be with `headerFields` (and other fields). I'm still confused as to why it's possible to add `semantics` to those, however given that it's possible to add to the root of the pass I'm good – still curious though 🙃

alexandercerutti:
> I'm still confused as to why it's possible to add semantics to those, however given that it's possible to add to the root of the pass I'm good – still curious though 🙃

You are in good company. I mean, as you reported, Apple states that it is possible to do so, but I honestly don't see the purpose of doing so. I mean, I find this behavior to be quite confusional. I'd put them all in the root.

Anyway, I was able to set semantics on a `headerField` without any issue:

immagine

buren:
> Anyway, I was able to set semantics on a headerField without any issue (picture of generated pass.json)

Nice, then this library follows the documented API as closely as possible - even when confusing 😅

alexandercerutti:
Oh yeah, definitely ahahah
I don't want to add further confusion or opinionated integrations.

PR #138, opened 2023-04-03
fix: transitType should be read when type is boardingPass

### LINKED ISSUE #137: Cannot proceed creating the pass because transitType is missing on your boardingPass.
To reproduce, just use the example here: https://github.com/alexandercerutti/passkit-generator/blob/11cfe99cccd27438edccc314aec044f6373f9fdf/examples/models/exampleBooking.pass/pass.json 

Load the model directly as shown in the readme:

```ts
	const pass = await PKPass.from({
		model: "./passModels/myFirstModel.pass",
		certificates: {
			wwdr,
			signerCert,
			signerKey,
			signerKeyPassphrase
		},
	});
```

So I added `transitType` in `boardingPass` in `pass.json`:

```ts
	"boardingPass": {
		"transitType": "PKTransitTypeGeneric"
	}
```

Still the same error.

So I printed out the PKPass instance:

image

The `boardingPass` got overwritten. I think this is caused by this line:

https://github.com/alexandercerutti/passkit-generator/blob/11cfe99cccd27438edccc314aec044f6373f9fdf/src/PKPass.ts#L387

I'm confused.

Songkeys:
## Description

This closes #137 

## Check relevant checkboxes

### DIFF
--- a/src/PKPass.ts
+++ b/src/PKPass.ts
@@ -556,6 +556,7 @@ export default class PKPass extends Bundle {
 				secondaryFields = [],
 				auxiliaryFields = [],
 				backFields = [],
+				transitType,
 			} = data[type] || {};
 
 			this.headerFields.push(...headerFields);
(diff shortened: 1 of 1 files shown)


alexandercerutti:
Does this work for you? I'll check later today. Thank you!

alexandercerutti (approved):
LGTM. Tested it. Thank you 🚀

PR #139, opened 2023-04-03
fix: complete pass field types for typescript

Songkeys:
Added more static types.

### DIFF
--- a/src/schemas/Field.ts
+++ b/src/schemas/Field.ts
@@ -1,26 +1,26 @@
 import Joi from "joi";
 import { Semantics } from "./Semantics";
 
-export type FieldDataDetectorType =
+export type PKDataDetectorType =
 	| "PKDataDetectorTypePhoneNumber"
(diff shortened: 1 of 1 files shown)


alexandercerutti (changes requested):
That's an interesting change.
I'd ask you to rename the fields to be compliant with the names.

So, for example:

`PKDataDetectorTypePhoneNumber` means `PKDataDetectionType` and not `FieldDataDetectorType`.

Songkeys:
Sure. PTAL.

In the meantime, I highly recommend [`zod` as a great alternative to `joi`](https://zod.dev/?id=joi). It offers static type inference, so you don't have to manually write them out.

Songkeys:
Interesting. I didn't know this since I've never used type checking to this level. Looks like they are going to introduce `switch` for this. Let wait and see. Great to learn something today!

alexandercerutti (approved):
LGTM! Thank you 🚀

PR #153, opened 2023-07-18
Added firebase example

Saim-Khan1:
## Description

Added a quick Firebase example, based on alexandercerutti#150, which initialises the firebase admin, downloads images stored in Firebase Cloud Storage to add to the pass (and deletes these once complete) and serves the pass directly as the response.

## Check relevant checkboxes

-   [ ] I've run tests (through `npm test`) and they passed


examples/firebase/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "node": "16"
+  },
+  "main": "index.js",
+  "dependencies": {
+    "axios": "^1.2.6",
+    "file-system": "^2.2.2",
+    "firebase-admin": "^10.0.2",
+    "firebase-functions": "^3.18.0",
+    "os": "^0.1.2",
+    "passkit-generator": "^3.1.7",
```
alexandercerutti, line 21: Do you mind putting `passkit-generator` as a `peerDependency`, just like I did with the other examples?

```
	"peerDependencies": {
		"passkit-generator": "latest"
	},
```

examples/firebase/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "functions",
+  "description": "Cloud Functions for Firebase",
+  "scripts": {
+    "serve": "firebase emulators:start --only functions",
```
alexandercerutti, line 5: Could you please add these scripts?

```
"service:link-pg": "cd ../.. && npm run build && npm link",
"preinstall": "npm run clear:deps",
"postinstall": "npm run service:link-pg && npm link passkit-generator",
"clear:deps": "rm -rf node_modules",
```

So that we can link the local passkit-generator version when installing.

examples/firebase/index.js
```diff
@@ -0,0 +1,179 @@
...
+            }
+            
+            // Downloading thumbnail and logo files from Firebase Storage and adding to pass
+            if (currentPassType == "generic" || currentPassType == "eventTicket") {
+                const thumbnailFile = request.body.thumbnailFile
+                const tempPath1 = path.join(os.tmpdir(), thumbnailFile)
+                try {
+                    await storageRef.file(`thumbnails/${thumbnailFile}`).download({destination: tempPath1})
+                } catch (error) {
+                    //
```
alexandercerutti, line 122: Please put at least a `console.error` here to print the error

examples/firebase/index.js
```diff
@@ -0,0 +1,179 @@
...
+                }
+                let buffer = Buffer.alloc(0);
+                try {
+                    buffer = fs.readFileSync(tempPath1)
+                } catch (error) {
+                    //
+                }
+                newPass.addBuffer("thumbnail.png", buffer)
+                newPass.addBuffer("thumbnail@2x.png", buffer)
```
alexandercerutti, line 131: I honestly do not understand what are you doing here: creating an empty buffer, and then ditching it for a file read from the FileSystem. Why?
I mean, okay for the fallback, but assigning an empty buffer would be better if done only in the catch.

Also, what I do not understand is that `tempPath1` is something that comes from a Firebase API... so using `fs` wouldn't make it crash or something when using this code on a Cloud Function, just as it happened on the scripapi's video?

examples/firebase/index.js
```diff
@@ -0,0 +1,179 @@
...
+            // Assigning certificates from certs folder (you will need to provide these yourself)
+            wwdr: fs.fs.readFileSync("./certs/wwdr.pem"),
+            signerCert: fs.fs.readFileSync("./certs/signerCert.pem"),
+            signerKey: fs.fs.readFileSync("./certs/signerKey.pem"),
+            signerKeyPassphrase: "PASSPHRASE"
+        }, 
+    },
+        {
+            authenticationToken: "AUTH_TOKEN",
+            webServiceURL: "https://us-central1-YOUR-FUNCTION.cloudfunctions.net/FUNCTION_NAME",
```
alexandercerutti, line 30: Is there a Firebase API that returns you, on runtime, the URL of the cloud function?

Perhaps this might be useless, as well as `authenticationToken`...

examples/firebase/index.js
```diff
@@ -0,0 +1,179 @@
...
+});
+var storageRef = admin.storage().bucket()
+
+exports.pass = functions.https.onRequest((request, response) => {
+    PKPass.from({
+        // Get relevant pass model from model folder (see passkit-generator/examples/models/)
+        model: `./model/${request.body.passType}.pass`,
+        certificates: {
+            // Assigning certificates from certs folder (you will need to provide these yourself)
+            wwdr: fs.fs.readFileSync("./certs/wwdr.pem"),
```
alexandercerutti, line 22: Wouldn't using `fs` on a cloud function make the cloud function crash? Maybe it is better to change paths here to `process.env` o something like this.

examples/firebase/index.js
```diff
@@ -0,0 +1,179 @@
...
+admin.initializeApp({
+    credential: admin.credential.cert(require("CERTIFICATE_PATH")),
+    storageBucket: "STORAGE_BUCKET_URL"
+});
+var storageRef = admin.storage().bucket()
+
+exports.pass = functions.https.onRequest((request, response) => {
+    PKPass.from({
+        // Get relevant pass model from model folder (see passkit-generator/examples/models/)
+        model: `./model/${request.body.passType}.pass`,
```
alexandercerutti, line 19: Is model being read correctly on a cloud function?

examples/firebase/index.js
```diff
@@ -0,0 +1,179 @@
...
+const { PKPass } = require("passkit-generator");
+const admin = require("firebase-admin");
+var fs = require('file-system');
+var path = require('path');
+var axios = require('axios');
+var os = require('os');
+
+// Firebase init
+admin.initializeApp({
+    credential: admin.credential.cert(require("CERTIFICATE_PATH")),
```
alexandercerutti, line 11: What is `CERTIFICATE_PATH` and what is `STORAGE_BUCKET_URL`? Should they be set from `process.env` perhaps?

alexandercerutti (changes requested):
Thank you very much for your contribution. This is wonderful, also I have several doubts although I'd rather have it in Typescript (but I guess I could convert it later).

What I noticed missing is a `README.md` file that explains how to use this example. If you want to write it, I can do it, but I need first to understand how everything in here works.

examples/firebase/package.json
Saim-Khan1: Sure, will do

examples/firebase/index.js
Saim-Khan1: Haha yep sorry, I'll go through and put those in

examples/firebase/index.js
Saim-Khan1: Yes these are probably both useless, looking back it looks like I just copied and pasted the function URL from the Firebase console and generated my own random auth token, so they can probably be removed!

examples/firebase/index.js
Saim-Khan1: It's not crashed from what I've seen? I've been generating passes using this function for a few months now, besides a few bits with the images not coming up on the pass (which I think is more to do with my frontend anyway), I haven't really had any issues with this. If you think there is a better way to do this though let me know, I'd be happy to change it

examples/firebase/index.js
Saim-Khan1: Yes that seems to work fine, essentially I have a folder which has everything in it, including a `model` folder containing some models I wrote based on the Apple Developer examples, each with the relevant name (e.g. boardingPass.pass, generic.pass etc.), and this entire folder gets deployed to Firebase. I think scriptpapi covers this in his video if I remember right? You might like have a look there if you'd like to see it in action.

examples/firebase/index.js
alexandercerutti: Yeah, remove them. I'll add them back later if I'll ever setup an example cloud function that supports passes update.

examples/firebase/index.js
Saim-Khan1: They're both to do with the firebase admin tools to access cloud storage – see https://firebase.google.com/docs/storage/admin/start
You get them both from the Firebase console, I had just pasted in the values and so when pushing this I simply replaced them with placeholders, sorry 😅
I'm not entirely sure what `process.env` is or how it works, but if that's the proper way it should be done please let me know and I'd be happy to try implementing :)

examples/firebase/index.js
Saim-Khan1: Ah I think I remember why I did that now, I think I declared the buffer and assigned it to be empty outside the try catch so it could then still be accessed in the `newPass.addBuffer()` lines? Unless there is a way to assign the buffer inside the try catch and still access it outside, or simply do the `newPass.addBuffer()` inside the try catch, in both cases? Let me know what you think is best, again really sorry if it's a trivial question!

alexandercerutti: The crash I refer to was about Scriptpapi writing on `fs` instead of Firebase and deploying the cloud function. It wasn't able to generate successfully a pass.

Let me ask you: you tried this on cloud or just locally? I think that `fs` is actually crashing but you are catching and losing the error if you are running this on the cloud.

In my opinion we should remove all the `fs` references and convert them to Firebase API.

examples/firebase/index.js
alexandercerutti: I'm wondering how it could read the files if Firebase cannot use FileSystem API 🤔

examples/firebase/index.js
alexandercerutti: Same here, I'm wondering how it could read the files if Firebase cannot use FileSystem API 🤔 What is `./model` relative to? Scriptpapi is actually using model, but I'm not sure what am I losing. Maybe `fs` works actually? But I honestly don't see where it is going to take the files from through `fs`.

examples/firebase/index.js
alexandercerutti: I've to check better how this works. I'll get back to you soon.

examples/firebase/index.js
Saim-Khan1: Hey, sorry for the delayed reply. I have tried this both locally using a firebase emulator, and the cloud, and I haven't seen any crashing (with the console logs put in), but you're right that might be a good idea generally. Do you have any idea what the firebase way to do it is?

examples/firebase/index.js
Saim-Khan1: Again I'm not sure, let me have another look at scriptpapi's video and try and figure out what's going on!

examples/firebase/index.js
alexandercerutti: I guess I was wrong about `fs`. Maybe it is okay reading... but not writing? I mean, here it is where scriptpapi shows the issue: https://youtu.be/rJZdPoXHtzI?t=2178.

So it is either sync writing or writing at all. Or maybe it is all about temp writing. I don't know.

I'd, maybe, use Cloud Storage for keeping files. If Cloud Functions work just like AWS Lambdas, I think that having files in the cloud folder, might increase the cold start function amount of time, as the "orchestrator" behind the scenes must clone or unzip or whatever all the files.

examples/firebase/index.js
alexandercerutti: As we were saying in another discussion above, maybe reading is fine but writing is illegal or something like this.

Anyway, maybe I'd not use the model as a folder in a Firebase cloud function, because it might mean a higher cold start time (see https://firebase.google.com/docs/functions/tips?hl=en#use_dependencies_wisely, even if it doesn't quote the File System).

I guess it is a tradeoff with the pricing. By looking [here](https://cloud.google.com/firestore/pricing?hl=en), for `eur3` (Europe region) - but maybe this is valid for all the regions - you have 50k free readings per day. This means that one might have a zip file and unzip it on runtime.

It depends on how many files you need to read, multiplied by the average number of passes you'll generate in a single day.

For the example purpose, using `fs` would be really enough, with a comment.

Or maybe two different examples should be created.

alexandercerutti:
I've partially reviewed the code you made right now:

1) Please note that if the model name, available at `request.body.passType` (which should be one of the models in the models folder, hence `exampleBooking` or `examplePass`), is checked against "boardingPass", it is never going to work. (not yet fixed).

2) It is better to have the certificates paths referencing the root folder, where `certificates` folder should be (as per the other examples)

3) A `firebase.json` is missing: I cannot run the function. I'm going to add one.

4) You didn't have prettier installed, so I had to change all the code indentation.

Since I've pushed on your branch, be sure to pull first if you want to make more changes

alexandercerutti:
Okay so, I was able to set up everything, deploy the function, and give the permission to be publicly accessed on the web (for future ref: https://cloud.google.com/functions/docs/securing/managing-access-iam?hl=it#console, with role `allUsers` to be inserted)

What I'm stuck on right now (I have to check better) is that the model as-is cannot be found on that path, because the model does not get uploaded.

Maybe we can copy the model folder on `predeploy` script so it gets deployed too (but then, we'd have a path mismatch between the emulator and deploying) or, otherwise, set it up on the bucket. WDYT @Saim-Khan1?

examples/firebase/index.js
alexandercerutti: I've removed `CERTIFICATE_PATH` and `STORAGE_BUCKET_URL` for a single `setupData.json` file

alexandercerutti:
Okay, I was able to overcome this block.
I think that the example is now ready! I'm going to merge it. If you have any doubts or questions, do not hesitate to ask them in a new discussion!

Thank you very much for your contribution @Saim-Khan1!

alexandercerutti (approved):

Saim-Khan1:
Hi @alexandercerutti, I was taking a little break, really sorry I missed all of this! Just having a quick read-through, looks like you managed to work through everything which is great, again really sorry I couldn't be of more help 😅
Great point about storing the models in cloud storage and the pricing tradeoff and things, I hadn't thought of that myself. I am yet to go through and have a look at how you went about doing things but I'm sure this will all be super useful in improving the backend for my app as well, so thank you so much again for all your help! I'll have a look through and let you know if I have any questions :)

PR #155, opened 2023-08-08
Fix Date to W3C string converter

JensSpanier:
## Description

Currently date conversion is not handled right if the timezone is different to UTC (Z).

**Example:**
I live in Germany (currently UTC+02:00). So if my local time is 12:00, it's 10:00 in UTC. The correct representation of this time as W3C (ISO 8601) string is `2023-08-08T10:00:00Z` or `2023-08-08T12:00:00+02:00`.

But if I call `dateToW3CString` with a date created with `new Date()` it's converted to `2023-08-08T12:00:00Z`. But thats 14:00 in my timezone. With `toISOString` it's correctly `2023-08-08T10:00:00.000Z`.

So instead of using this complex and incorrect way to convert a Date to a W3C (ISO 8601) string, we should use the built-in function `toISOString`.

## Check relevant checkboxes

### DIFF
--- a/specs/utils.spec.cjs
+++ b/specs/utils.spec.cjs
@@ -32,8 +32,8 @@ describe("Utils", () => {
 		});
 
 		it("should convert a Date object to a valid W3C date", () => {
-			expect(processDate(new Date(2020, 6, 1, 0, 0, 0, 0))).toBe(
-				"2020-07-01T00:00:00Z",
+			expect(processDate(new Date("2020-07-01T02:00+02:00"))).toBe(
+				"2020-07-01T00:00:00.000Z",
 			);
 		});
 	});
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -36,30 +36,7 @@ function dateToW3CString(date: Date) {
 		return undefined;
 	}
 
-	const paddedMonth = padMeTwo(date.getMonth() + 1);
-	const paddedDay = padMeTwo(date.getDate());
-	const paddedHour = padMeTwo(date.getHours());
-	const paddedMinutes = padMeTwo(date.getMinutes());
-	const paddedSeconds = padMeTwo(date.getSeconds());
-
-	/**
-	 * Date.prototype.getTimezoneOffset returns the timezone UTC offset in
-	 * minutes of the local machine.
-	 *
-	 * That value should then be used to calculate the effective timezone as
-	 * string, but still that would be related to the machine and not to the
-	 * specified date.
-	 *
-	 * For this reason we are completing date with "Z" TimeZoneDesignator (TZD)
-	 * to say it to use local timezone.
-	 *
-	 * In the future we might think to integrate another parameter to represent
-	 * a custom timezone.
-	 *
-	 * @see https://www.w3.org/TR/NOTE-datetime
-	 */
-
-	return `${date.getFullYear()}-${paddedMonth}-${paddedDay}T${paddedHour}:${paddedMinutes}:${paddedSeconds}Z`;
+	return date.toISOString();
 }
 
 function padMeTwo(original: string | number) {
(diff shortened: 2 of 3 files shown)


alexandercerutti:
Hey @JensSpanier, thank you for your PR! I'll take a deep look at it very soon, but at a first look, it looks already promising!

In the meanwhile I'd like to thank you for your contribution!

src/utils.ts
```diff
@@ -36,30 +36,7 @@ function dateToW3CString(date: Date) {
...
-	 * a custom timezone.
-	 *
-	 * @see https://www.w3.org/TR/NOTE-datetime
-	 */
-
-	return `${date.getFullYear()}-${paddedMonth}-${paddedDay}T${paddedHour}:${paddedMinutes}:${paddedSeconds}Z`;
+	return date.toISOString();
 }
 
 function padMeTwo(original: string | number) {
```
alexandercerutti, line 42: This function is useless now. You can remove it.

src/utils.ts
```diff
@@ -36,30 +36,7 @@ function dateToW3CString(date: Date) {
...
-	 * to say it to use local timezone.
-	 *
-	 * In the future we might think to integrate another parameter to represent
-	 * a custom timezone.
-	 *
-	 * @see https://www.w3.org/TR/NOTE-datetime
-	 */
-
-	return `${date.getFullYear()}-${paddedMonth}-${paddedDay}T${paddedHour}:${paddedMinutes}:${paddedSeconds}Z`;
+	return date.toISOString();
```
alexandercerutti, line 39: I'd still keep the URL that was available before (w3c).

specs/utils.spec.cjs
```diff
@@ -32,8 +32,8 @@ describe("Utils", () => {
 		});
 
 		it("should convert a Date object to a valid W3C date", () => {
-			expect(processDate(new Date(2020, 6, 1, 0, 0, 0, 0))).toBe(
-				"2020-07-01T00:00:00Z",
+			expect(processDate(new Date("2020-07-01T02:00+02:00"))).toBe(
```
alexandercerutti, line 35: I'm not really sure about these changes tests: wouldn't it be better to also have a test that uses the `Date` constructor with digits other than timestamps? What do you think? I mean, as a demonstration.

The same comment is valid for other tests you changed.

alexandercerutti (changes requested):

specs/utils.spec.cjs
JensSpanier: The problem with using `new Date(2020, 6, 1, 0, 0, 0, 0)` is, that it's interpreted in the local timezone.

> The parameter values are all evaluated against the local time zone, rather than UTC.

See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#individual_date_and_time_component_values).

So if I run `new Date(2020, 6, 1, 0, 0, 0, 0).toISOString()` on my computer, I get `2020-06-30T22:00:00.000Z`.

If someone in Sydney runs this, the result will be `2020-06-30T14:00:00.000Z`.

A possible solution would be to use [`Date.UTC()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC).

So the tests could look like this:
```ts
expect(processDate(new Date(Date.UTC(2020, 6, 1)))).toBe(
	"2020-07-01T00:00:00.000Z",
);
```

alexandercerutti: Okay cool. No need for further tests!

alexandercerutti (approved):
LGTM! Will merge this asap

alexandercerutti:
Released in v3.1.0 🎉 
Thank you very much @JensSpanier !

PR #245, opened 2025-05-27
Add support for stripColor

slootjes:
## Description

First of all, thanks for this library!

I was trying a lot of combinations and noticed that as soon as I set a strip image/buffer, the primary fields text is always rendered with a white color and did not follow the labelColor like I would expect. I stumped upon some places that mentioned "stripColor", tried it and it's indeed working. I couldn't get tests running on Windows 11 with Powershell unfortunately, I only got errors from jest (`SyntaxError: Cannot use import statement outside a module`).

Hope this change also can help others facing the same issue.

## Check relevant checkboxes

-   [ ] I've run tests (through `npm test`) and they passed

## Relevant information

Sources I could find about "stripColor":
- https://stackoverflow.com/a/31485785
- https://github.com/tinovyatkin/pass-js/pull/306
- https://help.passkit.com/en/articles/4179700-passkit-smartpass-links
- https://docs.passkit.io/common/templates/#operation/Templates_createTemplate

### DIFF
--- a/examples/firebase/functions/src/index.ts
+++ b/examples/firebase/functions/src/index.ts
@@ -20,6 +20,7 @@ interface RequestWithBody extends functions.Request {
 		textColor: string;
 		backgroundColor: string;
 		labelColor: string;
+		stripColor: string;
 		relevantDate?: string;
 		expiryDate?: string;
 		relevantLocationLat?: number;
--- a/src/schemas/index.ts
+++ b/src/schemas/index.ts
@@ -405,7 +405,7 @@ export type PassKindsProps = { [K in PassTypesProps]: PassProps[K] };
 
 export type PassColors = Pick<
 	OverridablePassProps,
-	"backgroundColor" | "foregroundColor" | "labelColor"
+	"backgroundColor" | "foregroundColor" | "labelColor" | "stripColor"
 >;
 
 export const PassPropsFromMethods = Joi.object<PassPropsFromMethods>({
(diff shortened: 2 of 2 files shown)


alexandercerutti:
Hey @slootjes, thank you for using passkit-generator and for contributing!

Jest issue is likely to be something related to how the process is run. I'm not sure how you are running the tests, but if you see the test commands, or the test command in `.vscode/settings.json`, you'll see that there's a flag to use ESM imports.

Talking about your change, thank you very much for the references.
I wonder if Apple forgot such property of if it is part of like a "private api". Not sure, but I've seen it as well from time to time in other places, but I've never added it.

Might sound stupid, but do you confirm it works with both RGB and HEX syntaxes? Just to be sure.

Let me know.

alexandercerutti:
The test failing in this PR is because the tests use some certificates and private key I created and set as env var in Github, but they are apparently not accessible from jobs run in PRs. Dunno why. So it fallbacks to file system paths, but the file is not available.

Do not worry. I saw the changes but nothing can be broken like this.

alexandercerutti (approved):
LGTM!

I'll merge it and release it later today. Thank you for your contribution, @slootjes!

If you could, also, leave a ⭐️ to the project, it would be great!

slootjes:
Amazing, thanks for your quick response. I've starred the repo and sent you a small donation for the hard work on this project, it's very useful :)

alexandercerutti:
Released in v**3.4.0** 🎉 Thank you very much again for improving passkit-generator @slootjes!

slootjes:
Awesome, happy to help and thanks for the quick merge ans release!

Issue #249, opened 2025-06-09
Supporting iOS 26 Changes


alexandercerutti:
Apple WWDC 25 just happened and today, Monday 09/06, Apple presented the changes for iOS 26. 

There are going to be:

 - Upcoming events for poster event tickets, with **a lot** of new properties
 - Upgraded boarding passes layout that uses semantics.

As we did last year with iOS 18 changes, we'll start _a hunt_ and start collecting details from the videos and from people that have the chance testing such changes.

The changes will be available in a new branch. I'll try to test them as soon as iOS 26 will be available in public beta (even if I can't ensure - I have only the personal iPhone).

> [!NOTE]
> The WWDC session can be viewed here: [What’s new in Wallet](https://developer.apple.com/videos/play/wwdc2025/202)

This post will get updated as soon new details will come out and in the first comment you will find all the properties that changed.

We are already lucky that the [documentation](https://developer.apple.com/documentation/walletpasses/pass) has **already** been updated, so it should be easier this time.

Join the hunt with me! Are you ready?

## Update

All the details have been recapped in the second comment, below, for you to have a rapid access. Comments still have screenshots you might want to see. Most relevant comments have been linked below.

TheEnbyperor:
From some quick experimenting, what's shown in the WWDC video is not enough to trigger the enhanced boarding pass. My guess would be something needs to go in `preferredStyleSchemes`, but this is not yet documented.

alexandercerutti:
@TheEnbyperor thanks for trying. You already tried on iOS 26 b1?
I'm trying to find all the possible things to check. If needed, I'll try to use reverse engineering against the OS as I did the last year.

This year Apple already did a great job by directly releasing the documentation.

TheEnbyperor:
Yep, I'm on the v26 developer beta.

alexandercerutti:
@TheEnbyperor In the documentation, the upgraded boarding passes are also named `airline boarding passes`.

With the `posterEventTickets` we had to put this key. If I'm right, we might have something like `airlineBoardingTicket` or `airlineBoardingPass` or something like that.

Do you mind checking?
Did you perhaps also check the `Console.app`? I remember last year the poster event tickets were sending logs where when providing invalid `preferredStyleSchemes` or something like that.

alexandercerutti:
Interesting. I'm going to set ipsw to download the b1 image. Let's go! :P

alexandercerutti:
[...]

PR #256, opened 2025-10-31
Optional venuePlaceID on UpcomingPassInformation

gpoitch:
## Description

This is re: https://github.com/alexandercerutti/passkit-generator/issues/249#issuecomment-3470813019
and version 3.5.3

I have working map and event guide tiles on the upcoming event details pages without providing a venuePlaceID, but that was made required in 3.5.3.

Other findings:
At least 1 backFields entry of any key/value is required to get the details pages.

## Check relevant checkboxes


alexandercerutti:
Released inside `v3.5.5`! Thanks for your contribution! 🚀

PR #270, opened 2026-06-27
Introducing iOS 27 support

### LINKED ISSUE #268: Supporting iOS 27 Changes
Hello folks!
Here we are for our yearly appointment to discuss and discover the new changes of Apple Wallet passes.

> [!NOTE]
> WWDC26 video can be found here: https://developer.apple.com/videos/play/wwdc2026/209/

The video already describes new features and [Apple Developer Documentation](https://developer.apple.com/documentation/walletpasses) seems to be already up to date (which is incredible!).

So there are a lot of changes:

- PosterGeneric

  They are just the new version of generic passes and makes them legacy (I might say). They do not use `preferredStyleSchemes` as we were used to think the past two years.

- New kinds of barcodes

  4 more types of barcodes have been introduced: `PKBarcodeFormatCode39`, `PKBarcodeFormatCodabar`, `PKBarcodeFormatEAN13` and `PKBarcodeFormatI2of5`.

  For what have been _Shaun_ (pun intended 👀) in the video, even iOS 18 Event Posters now support barcodes! (min: 4:35) **and they do not require NFC anymore if barcodes are provided**.

Image

- Featured Actions

  Up to two actions to be inserted under any (any?) kind of pass (to be sure, this should be tested). I didn't find yet any documentation about this, but I wonder if - in case of Poster Event Tickets and Semantic Boarding Passes, the actions will:

   - be the same
   - fallback to the original if user doesn't overwrite them
   - be entirely user-driven.

- A new pass designer MacOS Application

  This overlaps with what I designed and developed back in 2021, [pkvd](https://pkvd.app)

- A new format of templating from pass designer

  ~This will be very interesting, but this marks the fact that we might actually want to start supporting this in passkit-generator. It is a Protobuf file, so we'd have to figure out how to support it without impacting passkit-generator weight.~
  `.pkpasstemplate` are just [Apple Packages](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/DocumentPackages/DocumentPackages.html#//apple_ref/doc/uid/10000123i-CH106-SW1), so _directories_ that an application claim they can open. Nothing will change in passkit-generator for them.

Another point of view is: since Apple released [PassBuilder](https://github.com/apple/pass-builder) as Open Source Swift Package, another direction we might take is add support to the way they do this, like another implementation of passkit-generator (same package) that support the same methods.

As always, next comment will contain the more detailed recap of all the changes.
Let me know what you think of these changes.

~I'll leave this for a bit of time in order to gain feedback from people here and will start testing later in the summer (and perhaps reverse engineering `passd`, if needed, as I did the past two years).~

The update (v3.6.0) has been released!

alexandercerutti:
## Description

This pull request includes support to iOS 27 changes. Won't be merged officially until September, when we'll be near enough to the release of iOS 27 and therefore we'll be sure APIs won't change anymore.

However, alpha versions will be released with npm tag `alpha`, so you might want to try it with it in your package.json until `v3.6` won't land.

## Check relevant checkboxes


pingtumiinc:
Thanks for getting to the updates so quickly Alexander. I appreciate it. 

I'm in the process of testing but I could not find where I would add the footer field for the poster-generic type as described in the WWDC 2026 video "What’s new in Wallet" and displayed in this link https://developer.apple.com/documentation/walletpasses/creating-a-generic-pass. 

Also, this link seems to indicate that the featured actions can be customized. https://developer.apple.com/documentation/walletpasses/defining-the-metadata-of-your-wallet-pass 
such that the developer can add their custom title and icon. Does your implementation support custom featured actions? Below is an example.

```
pass.featuredActions = [
  {
    identifier: "openCollection",
    type: "place",
    icon: "map.circle.fill",
    title: "Open Collection",
    url: collectionURL
  },
  {
    identifier: "visitPingtumi",
    type: "membershipBenefits",
    icon: "sparkles.circle.fill",
    title: "Visit Pingtumi",
    url: pingtumiURL
  }
];
```

alexandercerutti:
Thanks for the feedback @pingtumiinc

~I have to check better the documents and video. I understood that the footer was always the same field.~ Yes, you are right. Footer fields seems to be a new kind of fields. I dunno why I was remembering they were already there.

For `featuredActions`, I didn't understand how a developer could define them. They talk about SFSymbols, but I thought it was a limited set, but maybe you are right. However, where do you read you can define a title?

pingtumiinc:
Apple docs say "Each featured action includes an icon, string, and action. You define your icon from available SF Symbols; be sure to pick one that accurately represents the action and that viewers can easily understand at a glance. Also, when choosing a symbol, prefer circular, filled icons. Your string is a brief call to action and your action label informs people what happens when they click the link; for example, your pass may redirect them to a website or the Maps app." I interpreted the sting to be the title. Apple Docs can be confusing at times.

pingtumiinc:
Also, I tried to update the pass kit generator but I when I run 
grep -R "footerFields" node_modules/passkit-generator -n
in the terminal, there is no mention of footerFields

alexandercerutti:
I've added `footerFields` to this PR but not yet released it. I will shortly (when I'll finish working :P).

> Apple docs say "Each featured action includes an icon, string, and action. You define your icon from available SF Symbols; be sure to pick one that accurately represents the action and that viewers can easily understand at a glance. Also, when choosing a symbol, prefer circular, filled icons. Your string is a brief call to action and your action label informs people what happens when they click the link; for example, your pass may redirect them to a website or the Maps app." I interpreted the sting to be the title. Apple Docs can be confusing at times.

I don't think there's a title at all. I think you can just change the icon and url, which may use a specific URI. Perhaps you can change only the `identifier`. I will make some more tests later.

Do you know where we can find SF Symbols?

alexandercerutti:
@pingtumiinc I tried to test different icons, title, action, other type. Nothing to do.
However, `type` and `identifier` are mandatory (`url` isn't, but pass doesn't do anything when you click it, ofc).

pingtumiinc:
SF Symbols is a separate Apple App for integrating icons into your app. you can learn more about it here https://developer.apple.com/sf-symbols/
You can use the pre-defined symbol name to access the symbol you want.

alexandercerutti:
Yes, I downloaded their app to copy the identifier and to import the image inside the `.pkpass` bundle, but no luck yet.

pingtumiinc:
What Apple says in the WWDC video and what they have on their website seem to conflict. The video does not indicate that the user can fully customize the featured actions but rather just provide a unique identifier, the action type and the url. If that's the case then the action type will determine what SF Symbol appears.

alexandercerutti:
I'm trying to write a topic on Apple Developer Forum but I can't send, it goes in HTTP 400, some kind of constraint validation failure, but not sure which.

pingtumiinc:
I was finally able to get my project up and running on Xcode27 and iOS27 using pass kit-generator. The new poster generic pass is displaying well. the only missing components are the logo, footer field and the featured actions.

This is the content of the .pkpas file

    "posterGeneric": {
        "headerFields": [
            {
                "key": "places",
                "label": "PLACES",
                "value": "8 places"
            }
        ],
        "primaryFields": [
            {
                "key": "title",
                "label": "PLACE COLLECTION",
                "value": "My Favorite Restaurants"
            }
        ],
        "secondaryFields": [
            {
                "key": "type",
                "label": "COLLECTION TYPE",
                "value": "Collection"
            }
        ],
        "auxiliaryFields": [
            {
                "key": "tagline",
                "label": "",
                "value": "Save it. Share it. Go."
            }
        ],
        "backFields": [
            {
                "key": "openCollectionURL",
                "label": "Open Collection",
                "value": "https://pingtumi.com/place-collection/8SRV64P7"
            },
            {
                "key": "pingtumiURL",
                "label": "Pingtumi",
                "value": "https://pingtumi.com/"
            },
            {
                "key": "collectionID",
                "label": "Collection ID",
                "value": "8SRV64P7"
            }
        ],
        "additionalInfoFields": []
    },
    "barcodes": [
        {
            "format": "PKBarcodeFormatQR",
            "message": "https://pingtumi.com/place-collection/8SRV64P7",
            "messageEncoding": "iso-8859-1",
            "altText": "Open Collection"
        }
    ],
    "featuredActions": [
        {
            "identifier": "openCollection",
            "type": "place",
            "url": "https://pingtumi.com/place-collection/8SRV64P7"
        },
        {
            "identifier": "visitPingtumi",
            "type": "membershipBenefits",
            "url": "https://pingtumi.com/"
        }
    ]
}

pingtumiinc:
Oops. I did not realize you had to add the pass to the wallet to view the featured actions. I added the pass but only one featured action appeared, the one with the identifier "visitPingtumi". The logo is still missing though as is the footer. I'll play around with it some more.

pingtumiinc:
I had some more time to play around with it. The featured actions seem to be working well although given the lack of customization, I'm not sure I'll be using them. Not sure why the logo is not appearing in the poster generic pass but appears in the generic pass.

alexandercerutti:
You mean this logo on left?
immagine

pingtumiinc:
Yes, that logo to the left of "MUSEUM".
Pingtumi Pass

pingtumiinc:
This is what the back/rear of the pass looks like. If the user taps Open Collection, the place collection will open. If you have an iPhone, just scan the QR Code and the place collection will open.

Pingtumi Pass Rear

alexandercerutti:
Let's take Pass Designer Pass as reference: Skull + "Museum" are a whole png logo (x2 and x3) themselves.

On Poster Generic, `primaryLogo` asset is used.
On classic generic ticket, `logo` is used.

immagine

Not sure which asset you are using.

pingtumiinc:
Thanks. I updated by adding "primaryLogo" to the asset file and it worked.

StewartM34:
Hi,

Thanks for working on this I’ve been looking at using posterGeneric and had a question around backwards compatibility.

Does the current implementation support passes being used on older versions of iOS?

From what I can ascertain from Apple’s documentation, their recommendation when using posterGeneric is to include both the new posterGeneric style and the existing generic style in the same pass.json. iOS 27 can then use the new poster layout, while iOS 26 and earlier can fall back to the generic layout.

From looking through this PR, it appears that passkit-generator still selects a single pass type, so I wasn’t sure whether it currently allows both generic and posterGeneric to coexist in the generated pass.

Is that already supported by this implementation, or is backwards compatibility something that still needs to be added?

alexandercerutti:
Hey @StewartM34, thanks for writing and using passkit-generator!

Yes, backward compatibility is supported, but as I was commenting on the linked issue, accessing both `posterGeneric` and `generic` might become a problem very easily as the `type` assumes only a specific type while now technically there may be more types to be supported (even if one is "legacy").

I'm in doubt on how to proceed with this, because either we add `"types"` property that supports both types or I'm not quite sure. I'm sure we are close to introduce a breaking change, but I'm not sure when.

StewartM34:
That makes sense, thank @alexandercerutti. So just to make sure I understand correctly: the generated pass can already contain both generic and posterGeneric for Apple’s backwards-compatible behaviour, but the current API/type model makes it difficult to independently access/configure both because it assumes a single pass type?

A types.generic / types.posterGeneric style API does sound quite logical to me, although I can see why that potentially becomes a breaking change.

alexandercerutti:
Yes, that's correct.
I'll have to think more about this before September comes and release a new alpha.

StewartM34:
Sweet Thank you!

pingtumiinc:
I was able to sole this problem by identifying the type of iOS device the user is using then have the sever generate appropriate json file.

 if #available(iOS 27.0, *) {
//Make a call to generate the poster generic json file
} else {
//Make a call to generate the generic json file
}

alexandercerutti:
@pingtumiinc I hate it, thanks! 😄 Are you using the alpha in production by any chance?

pingtumiinc:
Yes, I'm using the alpha in production. So far, it's working well.

StewartM34:
@pingtumiinc appreciate the suggestion but that workaround unfortunately doesn’t work for my use case. We attach the .pkpass directly to a confirmation email rather than serving it via an “Add to Wallet” link, so there’s no request-time opportunity to detect the device and generate the appropriate JSON. The pass has to work correctly regardless of which iOS version the recipient opens it on, whenever that is.

For scenarios like this, bundling both generic and posterGeneric in the same pass.json (as Apple’s docs describe) would be the ideal solution for me personally rather than relying on server-side device detection.

alexandercerutti:
This is a bit off topic but, @StewartM34, shouldn't it be better to have the "Add to Apple Wallet" button, as per Apple HIG, and make it download by request in order to avoid sending by-chance-useless bytes through email?

StewartM34:
> This is a bit off topic but, @StewartM34, shouldn't it be better to have the "Add to Apple Wallet" button, as per Apple HIG, and make it download by request in order to avoid sending by-chance-useless bytes through email?

Good question, and to be upfront about why we went with email over an on-demand "Add to Wallet" link - it wasn't really an efficiency choice, it came down to two things on our side:

QR/barcode generation. The barcode we embed is a live check-in token tied to a specific booking, generated server-side once the booking is confirmed. Serving that from a browser-clicked button means either minting a fresh valid token per request (which anyone holding the link could then use to pull a working ticket QR, without us ever confirming they're the actual guest) or baking a static one in ahead of time — which just puts us back at "generate once at confirmation," delivered differently.

Verification. Apple's pass-download protocol expects an Authorization: ApplePass ... header, which a plain link/button can't send - so a real on-demand button needs some session or short-lived token scheme to prove the requester is actually the ticket holder before handing over a working QR. That's meaningfully more surface area than "email it to the address already verified during booking."

So it wasn't a stylistic HIG preference so much as: on-demand generation would've required solving an auth problem we don't otherwise have, just to avoid an email attachment.

On the generic/posterGeneric conflict — since your concern is specifically that type assumes a single style and a full types.generic/types.posterGeneric API would be breaking, would an additive stopgap work? Leave the existing single-type setter exactly as-is (zero migration for current consumers), and add a separate opt-in method - e.g. .addFallbackType(type, fields) or .setPosterGeneric() - for anyone who wants both styles emitted in one pass. New capability, no breaking change, and it might let you ship something before September without fully resolving the type redesign. Curious if that holds up on your end or if there's an internal reason it can't work that way.

alexandercerutti:
@StewartM34

> On the generic/posterGeneric conflict — since your concern is specifically that type assumes a single style and a full types.generic/types.posterGeneric API would be breaking, would an additive stopgap work? Leave the existing single-type setter exactly as-is (zero migration for current consumers), and add a separate opt-in method - e.g. .addFallbackType(type, fields) or .setPosterGeneric() - for anyone who wants both styles emitted in one pass. New capability, no breaking change, and it might let you ship something before September without fully resolving the type redesign. Curious if that holds up on your end or if there's an internal reason it can't work that way.

The most probable solution is to introduce a `.types` and deprecate the `.type`.
If I'm right, next year Apple might do something similar with StoreCards and the next year, with Coupons (or vice versa), which means they might require more types as well.

StewartM34:
That seems sensible. If you want to let me know when you’ve got something ready to test, I’d be happy to give it a spin. Right now I’m kind of blocked and have to pick one or the other. Opting to build for the to be, rather than what used to be.

alexandercerutti:
Along with the property `types`, I'm considering the idea of introducing a class `PassType` that would contain and validate the fields. So in order to create a pass type, you will have to populate this class and put it inside the `types`.  What do you think of that, @pingtumiinc @StewartM34?

pingtumiinc:
My current implementation seems to be working fine so I'm not sure what
additional benefit it would bring me. I'm ok with it as long as it doesn't
break my current implementation.

StewartM34:
That works for me.

Would it make sense for the library to create the PassType objects internally, so we’d just work with something like pass.types.generic and pass.types.posterGeneric, rather than creating them ourselves?

alexandercerutti:
> My current implementation seems to be working fine so I'm not sure what
> 
> additional benefit it would bring me. I'm ok with it as long as it doesn't
> 
> break my current implementation.

@pingtumiinc I might argue your current production implementation is built upon an alpha version, which methods may change before the release 😅 I will do my best to not break what you already did but I can't promise. What I know is that I will keep (as deprecated) the signatures available today in the stable version.

@StewartM34 I was considering it. That would still imply to have the same structure beneath each one because of validation. 🤔

alexandercerutti:
Hello there again @StewartM34 @pingtumiinc, I've released a new alpha version, 3.6.0-alpha.2, that includes the following changes:

 - New export `PassType` you can use to create a type (`new PassType("boardingPass")`) and assign fields.
 - New getter `.types` that points to an array to which you can `push()` PassTypes.
 - Deprecated setter and getter `.type`. It setter now creates a new item inside `.types` and overrides the ones already available. Its getter now always returns the first item of `.types`.
 - Root `.primaryFields`, `.secondaryFields`, `.headerFields` and so on are deprecated. Same getters are available in `PassType`. Using them, will read the first element of `.types`.
 - Root `footerFields` getter has been removed for the one in `PassType`
 - `PassType` are freezable: once attached to a pass and the pass is emitted, they get frozen as well.
 - Deprecated `transitType` getter and setter. It can be set on `PassType` if the type is a `boardingPass`


Please, let me know what do you think of it!

StewartM34:
Hey @alexandercerutti 

Really appreciate your work on this. I’m currently away for the next couple of days, but once I’m back I will give it a whirl and report back.

pingtumiinc:
Thanks Alexander. I'll update in a few days and take a look.

pingtumiinc:
Hi Alexander:

I updated to version, 3.6.0-alpha.2 and everything works well. I was able to create passes in iOS 27 and non iOS 27 devices using a single call. I also noticed that you added the footer field which is great. Thanks for all your hard work.

alexandercerutti:
Hi @pingtumiinc, I forgot to reply to you. Thanks for letting me know!
I will perform the last checks tomorrow and likely release the stable version within the 09/09.

@StewartM34 were you able to test it?

StewartM34:
> Hi @pingtumiinc, I forgot to reply to you. Thanks for letting me know! I will perform the last checks tomorrow and likely release the stable version within the 09/09.
> 
> @StewartM34 were you able to test it?

Hey @alexandercerutti, I’ve only had a chance to test this briefly, but it looks like the generated .pkpass renders correctly. I opened the same .pkpass on both an iOS 27 beta device and a pre-iOS 27 device, and it renders appropriately on each - using the Poster generic ticket style on iOS 27, while falling back to the standard generic pass on pre-iOS 27 versions.

Issue #114, opened 2022-10-26
pass.getAsBuffer() fails on Lambda 


PCC94:
## Running OS

AWS Lambda

## Running Node Version

Node.js 16.x

## Description

Hi!

First at all, thank you for all your work!

To get into the point, i'm using your library on an AWS Lambda to generating .pkpass files and saving them on a S3 bucket.

Before I put it on lambda, i tried to do all the work on locale and it works well. As you can guess, there the troubles begans  when i put the same code on aws.

```
Duration: 3050.50 ms
Billed Duration: 3000 ms
Memory Size: 128 MB
Max Memory Used: 115 MB
Init Duration: 858.67 ms.	
Task timed out after 3.05 seconds
```
It seems like the the method `pass.getAsBuffer()` requests more memory and/or time than the lambda (as set) can offer, so i tried to extended these limits but it still don't work. I also tried to make the pass lighter removing all unecessary images, but nothing.

This is the function where the error occurs.

```javascript

  async createAndDownloadPass(username) {
        const s3 = new AWS.S3()
        
        const [signerCert, signerKey, wwdr] = await this.getCertificates() // no troubles here

        const pass = await PKPass.from({
            model: `${this.appleTemplatePath}/modelTemplate.pass`,
            certificates: {
                wwdr: wwdr,
                signerCert: signerCert,
                signerKey: signerKey,
                signerKeyPassphrase: process.env.SIGNER_KEY_PASSPHRASE
            },
        });
        pass.secondaryFields.push({
            "key": "username",
            "value": username
        })

        // no problems until here
        console.log('converting pass in buffer') // <--------------- Last log before it breaks
        
        const buffer = pass.getAsBuffer() // <---------------  the guilty?
       
        // never reached 
        const passName = `GeneratedPass-${Math.random()}.pkpass`;
        console.log(`uploading ${passName} on S3 bucket '${process.env.BUCKET_NAME}'`)
        
       // (i don't know even if it works)
        const {Location} = await s3 
            .upload({
                Bucket: ${process.env.BUCKET_NAME},
                Key: passName,
                ContentType: pass.mimeType,
                Body: buffer,
            })
            .promise();
        
        console.log(`${passName} Location: `, Location)
        return Location
    }

```
I think it's odd this operation costs so many resources anyway, so is there something wrong in the way i wrote the code?

Thank You

Paolo

## Expected behavior

`pass.getAsBuffer()` returns a buffer

## Steps to reproduce

[...]

PR #273, opened 2026-08-28
Speed up pass generation ~50x by removing four pure-JS hot spots

cuni0716:
## Description

Generating a pass is currently dominated by four independent hot spots, all of which
are avoidable without changing a single public signature. This series removes them
one at a time, smallest and safest first.

I hit this running Wallet passes in production: pass generation was the bottleneck in
the pipeline, and because the whole build is synchronous it blocked the event loop for
its entire duration, which is what actually hurts on a server.

Measured on a ten-file, ~450KB PNG bundle with an RSA-2048 key and two localizations.
These are Apple Silicon numbers, so the ratios are the transferable part, not the
absolutes:

| stage | mean | p95 | throughput | max event-loop block |
| --- | --- | --- | --- | --- |
| before | 62.77ms | 68.91ms | 16.3/s | 140.4ms |
| + signer key cache | 43.37ms | 47.36ms | 23.1/s | 74.1ms |
| + native SHA-1 | 38.69ms | 43.73ms | 26.1/s | 60.2ms |
| + linear zip writer | 19.59ms | 20.26ms | 51.8/s | 42.3ms |
| + native PKCS#7 | **1.24ms** | 1.46ms | **835.9/s** | **1.7ms** |

Roughly **50x** on per-pass latency and throughput, and ~80x on worst-case event-loop
blocking.

### The four commits

Each one is a separate commit that builds and passes the suite on its own, so they can
be reviewed — or merged — independently.

1. **Cache the parsed signer key.** The PEM was decrypted once per pass, re-running the
   key derivation function every time. Bounded to 32 entries, LRU-evicted, and keyed by
   a SHA-256 digest rather than the key material, so neither the PEM nor the passphrase
   is retained in a long-lived structure.
2. **Hash the manifest with `node:crypto`.** The manifest is a SHA-1 over every file in
   the bundle. Doing that in pure JS, via a latin1 string widened from the buffer,
   dominated hashing cost. latin1 is a lossless byte-to-character mapping, so the digest
   is unchanged.
3. **Replace `do-not-zip` with a linear zip writer.** `do-not-zip` does, per file:
   `fileData = [...fileData, 0x50, 0x4B, ..., ...dataBytes]` — rebuilding the entire
   archive in a plain JS array of numbers rather than a Buffer. ~450KB becomes ~450,000
   boxed numbers re-spread once per file: quadratic in bundle size, with heavy GC
   pressure. The replacement sizes the archive up front and writes it once into a single
   zero-filled Buffer, STORE-only (a pass is mostly PNG data that is already deflated).
4. **Sign with `node:crypto` instead of `node-forge`.** Pure-JS RSA was the single most
   expensive step — ~18.5ms per pass on its own. The SignedData is assembled with
   `asn1js` and signed with `crypto.sign()`.

## Check relevant checkboxes

-   [ ] I generated a working Apple Wallet Pass after the change

Rather than guessing at the three unchecked boxes:

- **Wallet pass on a device** — I generated complete `.pkpass` bundles and verified them
  with `openssl smime -verify` including full chain validation, an independent ASN.1
  walk that checks RSA via `node:crypto`, `unzip -t`, and by confirming every manifest
  digest matches the actual file bytes. But that was with *ephemeral* certificates, not
  real Apple signing material on a real device.
- **Examples** — not executed. Indirect evidence instead: the generated `lib/types` are
  byte-identical to `master` for every pre-existing declaration — `index.d.ts`,
  `PKPass.d.ts` and `Bundle.d.ts` all diff clean — so nothing the examples consume
  changed shape. The only additions are doc comments and a new internal `Zip.d.ts`,
  which `index.d.ts` does not export.
- **Breaking change** — the public API is not breaking, and that is asserted by a test
  rather than assumed. See *Runtime requirements* below for the one thing that does
  genuinely shift.

### The constraint that shaped this

`getAsBuffer()` is synchronous, and I wanted to keep it that way. A `pkijs` + WebCrypto
signer forces it to return a `Promise`, which is a breaking change for every caller and
a major version bump. `asn1js` encodes DER synchronously and `crypto.sign()` is
synchronous, so the whole path stays sync and this can land as a patch/minor.
`specs/output.spec.mjs` asserts the synchronous contract directly.

### Output is unchanged, and that is tested

The digests, file list and ordering pinned in `specs/output.spec.mjs` were captured from
an unmodified **3.5.7** before any of this work landed. Because the manifest is a SHA-1
of every other file, pinning it pins the exact bytes of `pass.json` and of each
`pass.strings` transitively, not merely their presence.

- **Zip** — byte-identical to `do-not-zip` across empty archives, empty entries, names
  longer than 255 bytes, multi-megabyte payloads and realistic bundles. Previously
  issued archives stay reproducible byte for byte.
- **Signature** — structurally identical to `node-forge`'s output. The
  `issuerAndSerialNumber`, both algorithm identifiers and the embedded certificate block
  are byte-identical; only `signingTime` and the signature bytes differ, as they do
  between any two runs of the *same* implementation.

Two deliberately conservative calls, both worth pushing back on if you disagree:

- **Authenticated attributes keep node-forge's insertion order** (contentType,
  messageDigest, signingTime) instead of being sorted as strict DER would require.
  Verifiers re-encode attributes exactly as parsed, and this is the order every pass
  issued by this library so far has carried, so re-sorting seemed like gratuitous risk
  for no gain.
- **File names stay latin1-encoded**, matching `do-not-zip`. This mangles non-ASCII
  names, but pass bundles are ASCII in practice, and fixing it is a behaviour change
  that does not belong in a performance PR. Happy to open a separate issue.

### Runtime requirements

The one thing that actually shifts, so flagging it explicitly:

- `zlib.crc32` is used when available (Node 20.15+) and falls back to a lookup table
  otherwise, so `engines: >=14.21.3` is **not** tightened.
- The signing path needs `crypto.createPrivateKey` and `crypto.sign`. On Node those are
  ancient. On **Cloudflare Workers**, the full `node:crypto` API landed in April 2025
  and requires only the `nodejs_compat` flag, which `examples/cloudflare-worker` already
  sets — so no `compatibility_date` bump should be needed. I was not able to execute
  that example to confirm, and it is the part I would most like a second opinion on.

### Dependencies

Runtime dependencies go from `do-not-zip, joi, node-forge, tslib` to
`asn1js, joi, tslib`, replacing `node-forge` with a considerably smaller tree.

`node-forge` and `do-not-zip` are retained as **devDependencies** and still earn their
place: forge generates ephemeral test certificates and verifies signatures independently
of the code that writes them, and `do-not-zip` backs the byte-compatibility assertions.
Producing with one stack and verifying with another is deliberate — a bug in the writer
cannot be cancelled out by a matching bug in the checker.

### Tests

101 passing. `specs/` gains:

- a dependency-free ZIP reader used as a test oracle;
- an independent PKCS#7 verifier that shares no code with the signing path;
- ephemeral certificate generation, so no real signing material is ever needed;
- coverage for tampered manifests, encrypted PKCS#8 keys with a `Bag Attributes`
  preamble, legacy encrypted PEMs, and wrong or missing passphrases.

Note the existing suite needs a WWDR certificate via the `WWDR` env var (or
`certificates/WWDR.pem`), which a fresh clone does not have. The new specs generate
their own certificates and need neither.

### On splitting this up

If four changes in one PR is too much at once, say the word and I will split it into
four stacked PRs — the commits are already separated for exactly that, and each builds
and passes the suite independently. The zip writer (commit 3) is the most self-contained
of them and has no crypto surface at all, so it may be the easiest starting point.

### DIFF
--- a/examples/cloudflare-worker/package.json
+++ b/examples/cloudflare-worker/package.json
@@ -8,7 +8,7 @@
 		"example": "pnpm wrangler dev"
 	},
 	"dependencies": {
-		"passkit-generator": "workspace:*"
+		"passkit-generator": "workspace:@jericommerce/passkit-generator@*"
 	},
 	"devDependencies": {
 		"@cloudflare/workers-types": "^4.20250109.0",
--- a/examples/firebase/functions/package.json
+++ b/examples/firebase/functions/package.json
@@ -18,10 +18,10 @@
 		"firebase-admin": "^13.0.2",
 		"firebase-functions": "^6.2.0",
 		"tslib": "^2.8.1",
-		"passkit-generator": "workspace:*"
+		"passkit-generator": "workspace:@jericommerce/passkit-generator@*"
 	},
 	"peerDependencies": {
-		"passkit-generator": "workspace:*"
+		"passkit-generator": "workspace:@jericommerce/passkit-generator@*"
 	},
 	"devDependencies": {
 		"firebase-functions-test": "^3.4.0",
--- a/examples/self-hosted/package.json
+++ b/examples/self-hosted/package.json
@@ -12,12 +12,12 @@
 		"example:debug": "pnpm tsx --inspect-brk src/index.ts"
 	},
 	"peerDependencies": {
-		"passkit-generator": "workspace:*"
+		"passkit-generator": "workspace:@jericommerce/passkit-generator@*"
 	},
 	"dependencies": {
 		"express": "^5.0.1",
 		"node-fetch": "^3.2.3",
-		"passkit-generator": "workspace:*",
+		"passkit-generator": "workspace:@jericommerce/passkit-generator@*",
 		"tslib": "^2.7.0"
 	},
 	"devDependencies": {
--- a/examples/serverless/package.json
+++ b/examples/serverless/package.json
@@ -14,7 +14,7 @@
 	"dependencies": {
 		"aws-sdk": "^2.1692.0",
 		"tslib": "^2.8.1",
-		"passkit-generator": "workspace:*"
+		"passkit-generator": "workspace:@jericommerce/passkit-generator@*"
 	},
 	"devDependencies": {
 		"@types/aws-lambda": "^8.10.147",
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -16,6 +16,16 @@ process.env.TZ = "UTC";
 
 export default {
 	moduleFileExtensions: ["js", "mjs", "cjs"],
+	/**
+	 * The specs import the package by its bare name, which used to resolve
+	 * through Node's self-reference on `name`. The package is published under
+	 * the @jericommerce scope, so that self-reference no longer matches. Map
+	 * the bare name back to the build to keep the specs identical to
+	 * upstream's.
+	 */
+	moduleNameMapper: {
+		"^passkit-generator$": "<rootDir>/lib/esm/index.js",
+	},
 	testEnvironment: "node",
 	testMatch: ["**/specs/**/*.spec.mjs"],
 	injectGlobals: false,
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
 		"Pass"
 	],
 	"dependencies": {
+		"asn1js": "^3.0.10",
 		"joi": "17.13.4",
-		"node-forge": "^1.4.0",
 		"tslib": "^2.7.0"
 	},
 	"engines": {
@@ -37,6 +37,7 @@
 		"do-not-zip": "^1.0.0",
 		"jest": "^29.7.0",
 		"jest-environment-node": "^29.7.0",
+		"node-forge": "^1.4.0",
 		"prettier": "^3.3.3",
 		"rimraf": "^4.4.1",
 		"tsconfig-to-dual-package": "^1.2.0",
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
 	"scripts": {
 		"build": "rm -rf lib && pnpm tsc -b tsconfig.esm.json tsconfig.cjs.json && pnpm build:dual",
 		"build:dual": "pnpm tsconfig-to-dual-package tsconfig.esm.json tsconfig.cjs.json",
+		"prepare": "rimraf lib && tsc -b tsconfig.esm.json tsconfig.cjs.json && tsconfig-to-dual-package tsconfig.esm.json tsconfig.cjs.json",
 		"build:all": "pnpm build && pnpm build:examples",
 		"prepublishOnly": "pnpm build && pnpm test",
 		"test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" pnpm jest -c jest.config.mjs --silent"
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-	"name": "passkit-generator",
+	"name": "@jericommerce/passkit-generator",
 	"version": "3.5.8",
 	"description": "The easiest way to generate custom Apple Wallet passes in Node.js",
 	"main": "lib/cjs/index.js",
@@ -8,15 +8,18 @@
 	"scripts": {
 		"build": "rm -rf lib && pnpm tsc -b tsconfig.esm.json tsconfig.cjs.json && pnpm build:dual",
 		"build:dual": "pnpm tsconfig-to-dual-package tsconfig.esm.json tsconfig.cjs.json",
-		"prepare": "rimraf lib && tsc -b tsconfig.esm.json tsconfig.cjs.json && tsconfig-to-dual-package tsconfig.esm.json tsconfig.cjs.json",
 		"build:all": "pnpm build && pnpm build:examples",
 		"prepublishOnly": "pnpm build && pnpm test",
 		"test": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" pnpm jest -c jest.config.mjs --silent"
 	},
 	"author": "Alexander Patrick Cerutti",
 	"license": "MIT",
-	"repository": "https://github.com/alexandercerutti/passkit-generator",
+	"repository": "https://github.com/cuni0716/passkit-generator",
 	"bugs": "https://github.com/alexandercerutti/passkit-generator/issues",
+	"publishConfig": {
+		"registry": "https://registry.npmjs.org",
+		"access": "public"
+	},
 	"keywords": [
 		"Apple",
 		"Passkit",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
 		"Pass"
 	],
 	"dependencies": {
-		"do-not-zip": "^1.0.0",
 		"joi": "17.13.4",
 		"node-forge": "^1.4.0",
 		"tslib": "^2.7.0"
@@ -33,9 +32,9 @@
 	},
 	"devDependencies": {
 		"@jest/globals": "^29.7.0",
-		"@types/do-not-zip": "^1.0.2",
 		"@types/node": "^16.11.26",
 		"@types/node-forge": "^1.3.11",
+		"do-not-zip": "^1.0.0",
 		"jest": "^29.7.0",
 		"jest-environment-node": "^29.7.0",
 		"prettier": "^3.3.3",
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
 		"Pass"
 	],
 	"dependencies": {
+		"asn1js": "^3.0.10",
 		"joi": "17.4.2",
-		"node-forge": "^1.3.2",
 		"tslib": "^2.7.0"
 	},
 	"engines": {
@@ -37,6 +37,7 @@
 		"do-not-zip": "^1.0.0",
 		"jest": "^29.7.0",
 		"jest-environment-node": "^29.7.0",
+		"node-forge": "^1.3.2",
 		"prettier": "^3.3.3",
 		"rimraf": "^4.4.1",
 		"tsconfig-to-dual-package": "^1.2.0",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
 		"Pass"
 	],
 	"dependencies": {
-		"do-not-zip": "^1.0.0",
 		"joi": "17.4.2",
 		"node-forge": "^1.3.2",
 		"tslib": "^2.7.0"
@@ -33,9 +32,9 @@
 	},
 	"devDependencies": {
 		"@jest/globals": "^29.7.0",
-		"@types/do-not-zip": "^1.0.2",
 		"@types/node": "^16.11.26",
 		"@types/node-forge": "^1.3.11",
+		"do-not-zip": "^1.0.0",
 		"jest": "^29.7.0",
 		"jest-environment-node": "^29.7.0",
 		"prettier": "^3.3.3",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -55,7 +55,7 @@ importers:
   examples/cloudflare-worker:
     dependencies:
       passkit-generator:
-        specifier: workspace:*
+        specifier: workspace:@jericommerce/passkit-generator@*
         version: link:../..
     devDependencies:
       '@cloudflare/workers-types':
@@ -80,7 +80,7 @@ importers:
         specifier: ^6.2.0
         version: 6.2.0(firebase-admin@13.0.2(encoding@0.1.13))
       passkit-generator:
-        specifier: workspace:*
+        specifier: workspace:@jericommerce/passkit-generator@*
         version: link:../../..
       tslib:
         specifier: ^2.8.1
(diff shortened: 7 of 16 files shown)


alexandercerutti:
Woah. This completely came out of the blue.
@cuni0716 first of all, thank you for the analysis and the pull request!

Let me answer you, improvement by improvement:

> Cache the parsed signer key. The PEM was decrypted once per pass, re-running the
key derivation function every time. Bounded to 32 entries, LRU-evicted, and keyed by
a SHA-256 digest rather than the key material, so neither the PEM nor the passphrase
is retained in a long-lived structure.

I'm not sure if this is something I want to be inside passkit-generator. I think it should perform one job and do it well. Adding a cache for the keys might be good **BUT** I think it should be an additional thing that stays on top of passkit-generator, not inside. Like a plugin or something. I think (so, no proof of this) most users use just one certificate. Having a way to inject a cache could be an interesting addition for the next major version. Not every project runs in a alive running system. Some of them run inside a Lambda / Cloud Function / Cloud Run / Cloudflare worker, so systems that shut down after a while of being inactive, for which the cache is useless.

> Hash the manifest with node:crypto. The manifest is a SHA-1 over every file in
the bundle. Doing that in pure JS, via a latin1 string widened from the buffer,
dominated hashing cost. latin1 is a lossless byte-to-character mapping, so the digest
is unchanged.

Hashing could be fine but signing with PKCS#7 is something that Node doesn't support.
I made some checks a few years ago with the native version for signing but it wasn't working. Likely, I was doing some mistakes. I will check this again.

> Replace do-not-zip with a linear zip writer. do-not-zip does, per file:
fileData = [...fileData, 0x50, 0x4B, ..., ...dataBytes] — rebuilding the entire
archive in a plain JS array of numbers rather than a Buffer. ~450KB becomes ~450,000
boxed numbers re-spread once per file: quadratic in bundle size, with heavy GC
pressure. The replacement sizes the archive up front and writes it once into a single
zero-filled Buffer, STORE-only (a pass is mostly PNG data that is already deflated).

I honestly see no benefit in this. The AI that wrote this, didn't get that perhaps if we are using a dependency is because we don't want to have the burden of handling zip format, so we prefer delegating it to a library that does one job and do that very well.

> Sign with node:crypto instead of node-forge. Pure-JS RSA was the single most
expensive step — ~18.5ms per pass on its own. The SignedData is assembled with
asn1js and signed with crypto.sign().

Linking to what said above... well, I think this might work.
We have to generate real passes because tests might not be covering this.

---

That said, things on this level, should be discussed first instead of opening a pull request directly.

I'm not going to close the PR yet because I'd like to discuss it with you.
What triggered these changes? What's your use case?

Thank you!


P.s. is 68.91ms too much to generate a pass? 🤔

cuni0716:
Thank you for taking the time to go through it improvement by improvement — and you're right that I should have opened a discussion first rather than arriving with four changes at once. Sorry about that.

Let me answer your question first, because it reframes the rest.

## What triggered this

We issue Apple Wallet passes in production on DigitalOcean App Platform (`apps-d-1vcpu-2gb` workers), RSA-2048 Apple certificates, ~430KB of PNG assets across a 10-file bundle. On that hardware:

- **160.9ms** per pass (p95 187ms), so ~6.2 passes/s per worker
- up to **263ms of fully synchronous event-loop blocking** per pass

The latency was survivable. The blocking was not. Because the whole build is synchronous, every pass stalls unrelated request handling on that worker — health checks, other routes, everything. A bulk issuance of 6000 passes took ~200s and needed a 5-instance fleet sitting at 66–91% CPU.

So this didn't start as "let's make a library faster". It started as "why does one worker fall over while issuing passes".

### > is 68.91ms too much to generate a pass? 🤔

On its own, honestly, no — for a single pass on request it's fine, and for most users it will never matter.

It matters when passes are issued in bulk, or when the process is also serving other traffic, because none of that time is yieldable. And I think it has already cost your users: in #114 someone's Lambda timed out at 3s/128MB, and their fix was to buy more memory and time. You wrote there:

> I know it might take a while due to signature generation that happens through `node-forge`. I do remember when I set up the Cloudflare Workers through community help, a performance report was kind of "alerting" that the heaviest part was the signature generation.

That diagnosis was correct. I measured `node-forge`'s PKCS#7 signing in isolation at **18.4ms per pass** versus **0.5ms** for the same RSA operation through `node:crypto`.

## Improvement by improvement

### The signer key cache — you're right, I'll drop it

Your single-responsibility argument is sound, and so is the serverless one. And the numbers agree with you: once signing goes native, the cache is worth **0.82ms** (2.08ms → 1.20ms), because `crypto.createPrivateKey` parses an encrypted PKCS#8 key in 0.54ms where `forge.decryptRsaPrivateKey` took 17ms. It was carrying real weight only because it was hiding a slow key parser.

Dropping it takes the series from 52x to 30x. That is a fine trade for keeping the library's scope clean, and an injectable cache in a next major is a better shape than what I proposed. I'll remove it.

### Hashing, and the PKCS#7 point

> Hashing could be fine but signing with PKCS#7 is something that Node doesn't support. I made some checks a few years ago with the native version for signing but it wasn't working.

You're right that Node has no PKCS#7 API, and I think that's exactly why the earlier attempt failed — it's the thing that tripped me up too. The approach here doesn't ask Node for PKCS#7 at all:

- **`asn1js`** builds the `SignedData` structure — certificates, `digestAlgorithms`, the detached `ContentInfo`, the `SignerInfo`, the authenticated attributes.
- **`node:crypto`** only computes the raw RSA signature over the DER-encoded authenticated attributes — one `crypto.sign("sha1", attrs, key)` call.

`crypto.createSign()` alone gives you a bare RSA signature, not a `SignedData`, which is what makes the naive attempt produce something Wallet rejects. Splitting "build the ASN.1" from "sign some bytes" is the whole trick.

Both halves are synchronous, which is the constraint I cared most about: `getAsBuffer()` keeps returning a `Buffer`, not a `Promise`. A `pkijs` + WebCrypto signer forces it async and makes this a major version.

I also checked the output rather than trusting it. Against `node-forge`'s signature, the `issuerAndSerialNumber`, both algorithm identifiers and the embedded certificate block are **byte-identical**; only `signingTime` and the signature bytes differ, as they do between any two runs of `node-forge` itself. I deliberately kept forge's *unsorted* authenticated-attribute order (contentType, messageDigest, signingTime) rather than sorting them as strict DER would want, because verifiers re-encode attributes exactly as parsed and that's the order every pass this library has ever issued carries.

### The zip writer

> I honestly see no benefit in this.

This one is on me — I presented it in the wrong order, so the benefit was invisible. Here's every subset measured on the same workload:

| variant | mean | throughput | vs 3.5.7 |
| --- | --- | --- | --- |
| 3.5.7 as-is | 62.77ms | 16/s | — |
| + native SHA-1 only | 54.00ms | 19/s | 1.2x |
| + native PKCS#7, **`do-not-zip` kept** | 18.86ms | 53/s | 3.3x |
| + linear zip writer | 2.08ms | 480/s | 30x |

The point is the last row. While `node-forge` dominates, the zip is invisible. Once signing is native — the change you're open to — **`do-not-zip` becomes 89% of everything that's left**, and the series stops at 3.3x instead of 30x. It isn't a micro-optimisation; it's the next bottleneck, and it only becomes visible after your preferred change lands.

The cause is one line — per file, `do-not-zip` does:

```js
fileData = [...fileData, 0x50, 0x4B, ..., ...dataBytes];
```

It rebuilds the entire archive on every file, in a plain JS array of numbers rather than a Buffer. A 450KB bundle becomes ~450,000 boxed numbers re-spread once per file. That's quadratic in bundle size, and most of the cost is GC.

On delegating to a dependency — I agree with the principle, and I'd rather have fixed it there. The obstacle is that **`do-not-zip`'s last release was November 2018**, with one maintainer and open issues unanswered since 2020. I'm happy to send the linear rewrite there and see, but I don't think it's a realistic path.

That said, I don't want this blocking the signature work, which is where the bulk of the win is. **Happy to split it into its own PR and let it be decided separately.**

### Generating real passes

> We have to generate real passes because tests might not be covering this.

Completely agree — this is the one thing my tests genuinely cannot prove. What I verified was: `openssl smime -verify` including full chain validation, an independent ASN.1 walk that checks the RSA signature via `node:crypto` and shares no code with the signing path, tampered manifests rejected by both, and `unzip -t`. But all with *ephemeral* certificates.

I'll validate with our real Apple certificates and confirm the passes install and open in Wallet on a physical device, and report back here with the results before you merge anything. If there's a specific pass type or device you'd want covered, tell me and I'll include it.

## Proposed way forward

Rather than push more commits at you, here's what I'd suggest — and I'd rather agree it with you first:

- **PR A — native SHA-1 + native PKCS#7 signing.** The main win (3.3x on its own), and the part you're most open to.
- **PR B — the linear zip writer.** Decided separately on its own merits, so it can't hold up PR A.
- **Dropped — the signer key cache.** Your call was right.

I'd also rebase onto 3.5.8 (the current conflict is only `package.json` / `pnpm-lock.yaml`), and rename two test passphrase literals that GitGuardian is flagging as secrets on your repo — they're just strings like `"the-right-one"` in a test asserting that a wrong passphrase throws, but there's no reason to leave the warning sitting there.

Happy to close this PR and open them fresh, or reshape this one — whichever is less noise for you. And if you'd rather move the discussion to an issue first, that's fine too.

Thanks again for engaging with it seriously rather than just closing it.

alexandercerutti:
Honestly, I think I will threat this whole PR and discussion as a <ins>set of suggestions</ins> that **I will NOT apply right now**, but that I may use to design and shape the next major version. I have to think singularly on each one of them.

> A bulk issuance of 6000 passes took ~200s and needed a 5-instance fleet sitting at 66–91% CPU.

I'm... a little bit concerned of this. I'm not into every user use case, but ideally an Apple Wallet pass should be generate lazily, so I'm not sure why the bulk.

Can you go deeper on the use case (business / project kind) and how passes are generated? Perhaps share the company name as well, just for statistics.

cuni0716:
That's completely fair — it's your library and your roadmap, and shaping a next major deliberately is a better outcome than merging four changes someone dropped on you unannounced. I'm glad the analysis is useful even if the code isn't what you take.

Happy to go deeper on the use case, since your question turns out to be the interesting part.

## The use case

Retail loyalty, on a multi-tenant platform. Store cards — points balance, tier, rewards — issued per merchant.

## On "why the bulk"

You're right that passes should be generated lazily, and that is what we do. "Bulk" was a bad word on my part: it made it sound like we pre-render passes into a bucket somewhere. We don't.

We implement the Apple Wallet web service the way the spec intends:

1. A customer's balance changes, or a merchant runs a campaign.
2. **We** send an APNs push to the devices holding that pass.
3. Each device responds by calling `GET /v1/passes/{passTypeIdentifier}/{serialNumber}`.
4. That callback is where the pass is generated — on demand, one device at a time.

So generation itself is lazy and on-demand: nothing is precomputed, nothing is stored, which is the advice you gave in #114 anyway.

What we don't control is the *concurrency*. We choose when to push, but once a merchant sends a campaign to their customer base, the devices decide when to answer — and in practice several thousand of them answer inside the same couple of minutes. The load isn't a batch job we scheduled; it's the shape of the response wave.

From the last 30 days of our telemetry:

- **~217,000 pass regenerations in a month**, each one a device callback
- **peak of 4,353 regenerations in a single minute** — roughly 73/second
- campaign hours around **12,200 regenerations/hour**, against a quiet-hour baseline under 300

That peak minute is the whole problem in one number. At the 6.2 passes/s per worker we measured, serving 73/s takes about 12 workers doing nothing else. With native signing it's under one. Same lazy generation, same response wave — the only variable is how long each callback ties up a worker.

And because the build is fully synchronous, a worker mid-generation isn't slow, it's *unavailable*: it can't serve the devices queued behind it, or anything else on that instance. That's the actual failure mode, and it's the same one the Lambda in #114 hit with a single pass.

So: the volume isn't why generation needs to be fast. The volume is just what made it impossible to ignore.

## Where we go from here

Practically, we'll run our fork in production from now on — no drama in that, it just unblocks us on the throughput side, and it means you're not under any pressure from us to take changes you're not comfortable with. If any of it turns out to be useful when you shape a v4, it's all there and you're welcome to any of it.

Shall I close this PR? It conflicts with 3.5.8 now and will only rot further sitting in your queue, and you've said you're not applying it as-is. If the reasoning is worth keeping I'm glad to write it up as a Discussion first — findings and measurements rather than a diff, each change with its own numbers and trade-offs — but I'm equally happy to just close it and stay out of your backlog. Your call.

And genuinely, thank you for passkit-generator. We built a real product on it, the Wallet spec is genuinely unpleasant to implement, and having a library that gets it right is the reason any of this existed to be profiled in the first place. Appreciate you engaging with the PR seriously rather than closing it.

alexandercerutti:
No, it is fine like this.
I'm sorry you have to run your fork just for this reason. I will do my best and tag you here once I'll figure out how the v4 will be.

Are you willing to be a beta tester for the v4?

cuni0716:
Yes, definitely — count me in for the v4 beta.

And no need to apologise about the fork. It's four patches sitting on top of your work, not a replacement for it, and having the library get the Wallet spec right is the part that actually mattered.

Tag me whenever there's something to try. Looking forward to seeing where you take it.

PR #275, opened 2026-09-15
Support placeIdentifier for place featured actions

bartolomeoadrian:
## Description

Fixes generation of `featuredActions` of `type: "place"`.

The `FeaturedAction` schema (added in #270 for iOS 27) hard-requires `url` on **every** action type and has no `placeIdentifier` key. But Apple's `place` action opens Maps to a location via a **`placeIdentifier`** (a Maps place ID), not a `url` — and a `place` action defined with a `url` does not work.

As a result, a valid place action such as:

```json
{
  "identifier": "action-1",
  "type": "place",
  "placeIdentifier": "I87A8F52C8A209A35"
}
```

…could never be produced. During validation (`stripUnknown: true`) the unknown `placeIdentifier` is dropped, the now-missing required `url` throws, and `filterValid` silently swallows the error and discards the whole action — so it just disappears from `pass.json` with only a `console.warn`.

### Change

Made the schema type-aware in `src/schemas/FeaturedAction.ts`:

- Added `placeIdentifier?: string` to the `FeaturedAction` interface and made `url?` optional.
- Joi conditional on `type`:
  - `type: "place"` → **`placeIdentifier` required**, `url` optional
  - any other type → **`url` required** (unchanged behavior)

Non-place actions are completely unaffected.

## Check relevant checkboxes

## Relevant information

- Added test coverage in `specs/PKPass.spec.mjs` (new `featuredActions` describe block): a place action defined via `placeIdentifier` is preserved and lands in the generated `pass.json`; a place action missing `placeIdentifier` is dropped; a non-place action via `url` is kept; a non-place action missing `url` is dropped; and the two-action cap still holds.
- **Verified against real Wallet behavior.** I built two Pass Designer templates: a `place` featured action defined via `url`, and one defined via `placeIdentifier`.
  - A `place` action defined with a **`url`** does **not render on device** (iOS 27), and Pass Designer shows it as a **broken action** — *"This action's URL or Maps Place ID is invalid."*
  - A `place` action defined with a **`placeIdentifier`** renders correctly ("Get Directions / Open in Maps").

  This confirms `place` actions must be defined through `placeIdentifier`, which is what this change enforces. (Pass Designer template bundles and the screenshot are attached below.)
- Reference: iOS 27 featured actions ([WWDC26 "What's new in Wallet"](https://developer.apple.com/videos/play/wwdc2026/209/)).

[FeaturedAction-place-placeIdentifier.pkpasstemplate.zip](https://github.com/user-attachments/files/32237780/FeaturedAction-place-placeIdentifier.pkpasstemplate.zip)
[FeaturedAction-place-url.pkpasstemplate.zip](https://github.com/user-attachments/files/32237782/FeaturedAction-place-url.pkpasstemplate.zip)

Result of `FeaturedAction-place-url.pkpasstemplate`:

passdesigner-place-url-invalid


alexandercerutti:
Hi @bartolomeoadrian, thanks for this PR. I will address it as soon as possible (when I will not be ill and/or find some time).

FYI, I really dislike commits co-authored by Claude, so you are still in time to rebase them and force-push them. Thanks.

bartolomeoadrian:
> Hi @bartolomeoadrian, thanks for this PR. I will address it as soon as possible (when I will not be ill and/or find some time).
> 
> FYI, I really dislike commits co-authored by Claude, so you are still in time to rebase them and force-push them. Thanks.

Thanks! I have rebased the branch. Please let me know if I can make any modifications here, I would very much like to merge this PR and become a contributor.

src/schemas/FeaturedAction.ts
alexandercerutti, line 1: I think that at this point, we might make `FeaturedAction` something like a discriminated union like: if we have a `type: "place"`, require `placeIdentifier`. Otherwise, use `url`. The same would apply to Joi scheme.

Then, I think we might do a little bit more than this. Pass Designer exports this for Calls.

```
      "identifier" : "Test aall",
      "type" : "call",
      "url" : "tel:\/\/+393496847717"
```

So we might enforce the `tel://` protocol. I tried to remove it and use a different protocol but it always appends the `tel://`. So I wonder if it is a limitation of Pass Designer or an actual necessity.

What do you think of this?

src/schemas/FeaturedAction.ts
```diff
@@ -88,5 +97,16 @@ export const FeaturedAction = Joi.object<FeaturedAction>().keys({
 			"featuredActionType",
 		)
 		.required(),
-	url: Joi.string().uri().required(),
+	url: Joi.string()
+		.uri()
+		.when("type", {
+			is: "place",
+			then: Joi.optional(),
```
alexandercerutti: Other than the file comment I left above, if `type === "place"`, I think `url` should be `Joi.string().forbidden()`

alexandercerutti: Same, of course, applies below to the `placeIdentifier`.

alexandercerutti (changes requested):

src/schemas/FeaturedAction.ts
bartolomeoadrian: Done

src/schemas/FeaturedAction.ts
bartolomeoadrian: I’d avoid enforcing `tel://` just in case Apple rolls out other protocol options. It’s probably best to let developers decide what works for them.

alexandercerutti:
I wonder if we can add a test to check Joi behavior. I mean that we omit a field like that.

bartolomeoadrian:
> I wonder if we can add a test to check Joi behavior. I mean that we omit a field like that.

Done

alexandercerutti (approved):
LGTM!

